PR #1, opened 2021-07-14
Initial implementation

turkenh:
Adds initial implementation for provider-kubernetes.

Provider Kubernetes enables deployment and management of arbitrary Kubernetes objects on Kubernetes clusters typically provisioned by Crossplane. In other words, it is similar to [provider-helm](https://github.com/crossplane-contrib/provider-helm) but for Kubernetes resources rather than helm packages.

For example, the following `Object` resource would create a namespace on the Kubernetes cluster configured with `providerConfigRef`.  

```
kind: Object
metadata:
  name: team-a-namespace
spec:
  forProvider:
    manifest:
      apiVersion: v1
      kind: Namespace
      metadata:
        # name in manifest is optional and defaults to Object name
        # name: some-other-name
        labels:
          example: "true"
  providerConfigRef:
    name: kubernetes-provider
```


internal/controller/object/object.go
```diff
@@ -0,0 +1,310 @@
...
+const (
+	errTrackPCUsage = "cannot track ProviderConfig usage"
+	errGetPC        = "cannot get ProviderConfig"
+	errGetCreds     = "cannot get credentials"
+	errGetObject    = "cannot get object"
+	errCreateObject = "cannot create object"
+	errApplyObject  = "cannot apply object"
+	errDeleteObject = "cannot delete object"
+
+	errNotKubernetesObject      = "managed resource is not a Object custom resource"
```
hasheddan, line 56: the biggest nitpick of all time :)
```suggestion
	errNotKubernetesObject      = "managed resource is not an Object custom resource"
```

internal/controller/object/object.go
```diff
@@ -0,0 +1,310 @@
...
+	c.logger.Debug("Updating", "resource", cr)
+
+	obj, err := getDesired(cr)
+	if err != nil {
+		return managed.ExternalUpdate{}, err
+	}
+
+	obj.SetAnnotations(map[string]string{
+		v1.LastAppliedConfigAnnotation: string(cr.Spec.ForProvider.Manifest.Raw),
+	})
```
hasheddan, line 250: @turkenh does this cause any conflict with k8s trying to set this / will k8s not do this automatically?

hasheddan (commented):
This is awesome @turkenh! All comments are minor :)

internal/controller/object/object.go
turkenh: k8s does not set this automatically rather clients like `kubectl apply` set this. I also noticed ArgoCD also sets the same.
So, no conflict as long as someone kubectl apply's on the object, which would be reverted back by the provider-kubernetes controller.

turkenh:
> This is awesome @turkenh! All comments are minor :)

thanks @hasheddan, your comments should be resolved now.

hasheddan (approved):

morningspace (commented):
Just a general comment regarding the naming: `Object`. Would it be better if we change it to something else such as `Managed` or `Managable`. The reason is that, `Object` is too generic, where in this case, the object is essentially a wrapper or a representation of the raw Kubernetes resource as a managed resource in the Crossplane world. :-)

turkenh:
> Just a general comment regarding the naming: `Object`. Would it be better if we change it to something else such as `Managed` or `Managable`. The reason is that, `Object` is too generic, where in this case, the object is essentially a wrapper or a representation of the raw Kubernetes resource as a managed resource in the Crossplane world. :-)

Thanks for your suggestion @morningspace!

I am not a 100% sure regarding `Object` either. However, I still believe it represents best what is **manages**, i.e. Kubernetes objects. `Managed` (or `Managable`) would also be too generic and also confusing in Crossplane world since it represents a category (`kubectl get managed` returns all crossplane managed resources).

Another alternative that I considered was `Resource` since we usually mention them as arbitrary Kubernetes **resources**, however, check through Kubernetes documentation I noticed `Object` better represents what we are referring here:

See [object](https://kubernetes.io/search/?q=object) vs [resource](https://kubernetes.io/search/?q=resource).

P.S.: Full API representation becomes `objects.kubernetes.crossplane.io`

morningspace:
> > Just a general comment regarding the naming: `Object`. Would it be better if we change it to something else such as `Managed` or `Managable`. The reason is that, `Object` is too generic, where in this case, the object is essentially a wrapper or a representation of the raw Kubernetes resource as a managed resource in the Crossplane world. :-)
> 
> Thanks for your suggestion @morningspace!
> 
> I am not a 100% sure regarding `Object` either. However, I still believe it represents best what is **manages**, i.e. Kubernetes objects. `Managed` (or `Managable`) would also be too generic and also confusing in Crossplane world since it represents a category (`kubectl get managed` returns all crossplane managed resources).
> 
> Another alternative that I considered was `Resource` since we usually mention them as arbitrary Kubernetes **resources**, however, checking through Kubernetes documentation I noticed `Object` better represents what we are referring here:
> 
> See [object](https://kubernetes.io/search/?q=object) vs [resource](https://kubernetes.io/search/?q=resource).
> 
> P.S.: Full API representation becomes `objects.kubernetes.crossplane.io`

Understood. Thanks for clarification. Naming is always a tricky work :-)

AaronME (approved):
The local-cluster example worked for me deploying from one kind cluster to another.

I was not able to get the out-of-cluster provider to work with the composition, but do not think that's a factor of the code.

One tiny note on repository.

Makefile
```diff
@@ -0,0 +1,102 @@
...
+# ====================================================================================
+# Setup Kubernetes tools
+KIND_VERSION = v0.11.1
+USE_HELM3 = true
+-include build/makelib/k8s_tools.mk
+
+# ====================================================================================
+# Setup Images
+
+DOCKER_REGISTRY = turkenh
```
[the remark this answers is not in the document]
turkenh: ah, good catch, thx!

PR #19, opened 2021-12-07
Add controller reference to controlled resources

### LINKED ISSUE #18: OwnerReference not added to kubernetes resources
I created a composition as attached that is just supposed to create a namespace when a `project.example.com` resource is created.

The resulting namespace is created as expected, however it doesn't have an owner reference set. That means when I delete my `project.example.com` resource, the provisioned namespace remains on the cluster.

It should have the owner reference set so that it gets cleaned up by the kubernetes garbage collector once the composite `objects.kubernetes.crossplane.io` resource is gone.

```yaml
apiVersion: apiextensions.crossplane.io/v1
kind: Composition
metadata:
  name: compositeprojects.example.com
  labels:
    provider: me
spec:
  compositeTypeRef:
    apiVersion: example.com/v1alpha1
    kind: CompositeProject
  patchSets:
  - name: metadata
    patches:
    - fromFieldPath: metadata.labels
  resources:
    - name: namespace
      base:
        kind: Object
        spec:
          forProvider:
            manifest:
              apiVersion: v1
              kind: Namespace
            metadata:
              name: ""
          providerConfigRef:
            name: kubernetes-provider
      patches:
        - type: PatchSet
          patchSetName: metadata
        - fromFieldPath: "spec.claimRef.name"
          toFieldPath: "spec.forProvider.manifest.metadata.annotations[paas.getcruise.com/repos]"
        - fromFieldPath: "spec.claimRef.name"
          toFieldPath: "spec.forProvider.manifest.metadata.name"
```

jonnylangefeld:
The `Object` resource provisions other kubernetes resources. However those resulting resources do not have an owner reference set, which leaves them abandoned on the cluster once the `Object` resource gets deleted.

This properly sets the owner reference on the controlled resource using controllerutils.SetControllerReference.

A *runtime.Scheme is needed for the function which can be optained via external.client.Scheme(). Rather than passing that *runtime.Scheme as parameter, I made `getDesired` a receiver function of `external`, so it would have direct access to it. It was a private function anyway, so no breaking changes

Fixes #18 

I have:

### DIFF
--- a/internal/controller/object/object.go
+++ b/internal/controller/object/object.go
@@ -32,6 +32,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/event"
@@ -191,7 +192,7 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 
 	c.logger.Debug("Observing", "resource", cr)
 
-	desired, err := getDesired(cr)
+	desired, err := c.getDesired(cr)
 	if err != nil {
 		return managed.ExternalObservation{}, err
 	}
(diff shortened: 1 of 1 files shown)


jonnylangefeld:
Hmm I don't think this solves it as it's not possible to have a namespaced owner on a cluster resource. I'm getting this error:
`cluster-scoped resource must not have a namespace-scoped owner, owner's namespace test-namespace`

jonnylangefeld:
I don't think this fix is needed anyway: https://github.com/crossplane-contrib/provider-kubernetes/issues/18#issuecomment-987453247

PR #21, opened 2022-01-04
init commit for feature mgmt policy and reference

### LINKED ISSUE #13: Design proposal to enhance provider kubernetes with new features

This issue is aimed to track all discussions around the design proposal documented in https://github.com/morningspace/capabilities-shim/blob/main/docs/enhanced-provider-k8s.md. It adds below features to provider-kubernetes.
- Allow user to define resource management policy to instruct the provider how to manage Kubernetes resources in a fine-grained manner.
- Allow user to define resource references for an `Object` as its dependencies to retrieve values from these dependent resources at runtime and guarantee the resource rendering in a specified order. (This implements the [Generic references design](https://github.com/crossplane/crossplane/pull/2385) at provider level).

There is a reference implementation that adds the above features at: https://github.com/morningspace/provider-kubernetes/tree/capabilities-shim, which is a repository forked from here.

By implementing the above features, user can:
- Manage arbitrary Kubernetes resources using provider-kubernetes in a much fine-grained manner.
- Patch resource with values by resolving from its dependent resources.
- Guarantee both install and uninstall in specified order.

morningspace:
Signed-off-by: Ying Mo <morningspace@yahoo.com>

Fixes #13 

I have:


morningspace:
Hi @turkenh This is the initial commit for an early review. The feature has been manually tested at local and I was adding test cases to catch up the code coverage, will submit a second commit including the test cases soon.

morningspace:
Added test cases and now the coverage has been increased +26.46% :-) https://codecov.io/github/crossplane-contrib/provider-kubernetes/commit/d7bcc5a8e1aff6bad593147b0f2b95fd4af31405

apis/object/v1alpha1/types.go
```diff
@@ -74,3 +127,46 @@ type ObjectList struct {
...
+	}
+
+	toMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(to)
+	if err != nil {
+		return err
+	}
+	if err := fieldpath.Pave(toMap).SetValue(path, value); err != nil {
+		return err
+	}
+	return runtime.DefaultUnstructuredConverter.FromUnstructured(toMap, to)
```
turkenh, line 171: Seems like we can just use existing fieldpath functions, if I am not missing something.

```suggestion
	paved, err := fieldpath.PaveObject(to)
	if err != nil {
		return err
	}
	return paved.SetValue(path, value)
```

internal/controller/object/object.go
```diff
@@ -65,6 +65,13 @@ const (
 	errGetLastApplied          = "cannot get last applied"
 	errUnmarshalTemplate       = "cannot unmarshal template"
 	errFailedToMarshalExisting = "cannot marshal existing resource"
+
+	errGetReferencedResource       = "cannot get referenced resource"
+	errPatchFromReferencedResource = "cannot patch from referenced resource"
+	errResolveResourceReferences   = "cannot resolve resource references"
+
+	errFailedToAddFinalizer    = "Failed to add finalizer to referenced resource"
+	errFailedToRemoveFinalizer = "Failed to remove finalizer to referenced resource"
```
turkenh, line 74: error messages should start with lowercase.

nitpick: lets use `cannot` instead of `failed to` just for consistency

```suggestion
	errFailedToAddFinalizer    = "cannot add finalizer to referenced resource"
	errFailedToRemoveFinalizer = "cannot remove finalizer to referenced resource"
```

internal/controller/object/object.go
```diff
@@ -246,6 +243,12 @@ func (c *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 	}
 
 	c.logger.Debug("Creating", "resource", cr)
+
+	if policy := cr.Spec.ManagementPolicy; policy == v1alpha1.ObserveDelete || policy == v1alpha1.Observe {
```
turkenh, line 247: Having a simple function to decide whether an action is allowed or not would improve readability, something like: (prefer defining an enum instead of string as input though)

```
if !isAllowedTo("Create") {

}
```

apis/object/v1alpha1/types.go
```diff
@@ -18,10 +18,61 @@ package v1alpha1
...
+)
+
+// FromObject refers to an object by Name, Kind, APIVersion, etc. It is used
+// to reference other Object or arbitrary Kubernetes resource which is either
+// cluster or namespace scoped.
+type FromObject struct {
+	// APIVersion of the referenced object.
+	APIVersion string `json:"apiVersion"`
+	// Kind of the referenced object.
+	Kind string `json:"kind"`
```
turkenh, line 51: Should we just _default_ to a Kubernetes resource by defining default APIVersion and Kind?

apis/object/v1alpha1/types.go
```diff
@@ -18,10 +18,61 @@ package v1alpha1
...
+	// Name of the referenced object.
+	Name string `json:"name"`
```
turkenh, line 53: I believe at some point we will need to add support for label selectors similar to existing crossplane references with selectors, e.g. `vpcIdSelector`, but it is totally ok (even preferable) to postpone it to later PR.

internal/controller/object/object.go
```diff
@@ -340,3 +354,160 @@ func setObserved(obj *v1alpha1.Object, observed *unstructured.Unstructured) erro
...
+		var kubeClient client.Client
+
+		if ref.FromObject.APIVersion == obj.APIVersion && ref.FromObject.Kind == obj.Kind {
+			// Object resource handled by provider on local cluster
+			kubeClient = c.localClient
+		} else {
+			// Other resource usually handled on remote cluster
+			// But could also be handled on local cluster
+			kubeClient = c.client
+		}
```
turkenh, line 375: IIUC, this seems a bit problematic to me. 
- How could we decide to use localclient by just checking if `APIVersion` and `Kind` matches for object itself and it references? 

My understanding so far is, we will always use localclient for dependencies and references. If we need to reference a remote resource, e.g. a K8S CR, we will need to create a local Object CR tracking it with Observe/ObserveDelete policy and resolve references from it.

Am I missing something, could you elaborate a bit?

internal/controller/object/object.go
```diff
@@ -340,3 +354,160 @@ func setObserved(obj *v1alpha1.Object, observed *unstructured.Unstructured) erro
...
+			return errors.Wrap(err, errGetReferencedResource)
+		}
+
+		// Patch fields if any
+		if err := ref.ApplyFromFieldPathPatch(res, obj); err != nil {
+			return errors.Wrap(err, errPatchFromReferencedResource)
+		}
+
+		// Add finalizer if not exists
+		finalizer := "object/" + string(obj.UID)
```
turkenh, line 396: a more descriptive string would be better, e.g. `kubernetes.crossplane.io/referred-by-object: <obj.UID>`

internal/controller/object/object.go
```diff
@@ -295,6 +303,12 @@ func (c *external) Delete(ctx context.Context, mg resource.Managed) error {
 	}
 
 	c.logger.Debug("Deleting", "resource", cr)
+
+	if policy := cr.Spec.ManagementPolicy; policy == v1alpha1.ObserveCreateUpdate || policy == v1alpha1.Observe {
+		c.logger.Debug("External resource should not be deleted by provider, skip updating.")
```
turkenh, line 308:
```suggestion
		c.logger.Debug("External resource should not be deleted by provider, skip deleting.")
```

internal/controller/object/object.go
```diff
@@ -340,3 +354,160 @@ func setObserved(obj *v1alpha1.Object, observed *unstructured.Unstructured) erro
...
+			c.logger.Debug(errGetReferencedResource, "error", err)
+			continue
+		}
+
+		// Remove finalizer if exists
+		finalizer := "object/" + string(obj.UID)
+		if meta.FinalizerExists(res, finalizer) {
+			meta.RemoveFinalizer(res, finalizer)
+			if err := kubeClient.Update(ctx, res); err != nil {
+				c.logger.Debug(errFailedToRemoveFinalizer, "error", err)
```
turkenh, line 444: I don't think it is a good idea to ignore error and continue, we should return it. This could cause unexpected issues which would require manual intervention. But returning error would trigger a retry and it could fix issue in the next tries, e.g. in case of a transient connectivity problem.

I would already suggest to call `removeReferenceFinalizers()` in `Delete` method instead of `isNotFound()` which would give use the possibility to return an error.

internal/controller/object/object.go
```diff
@@ -340,3 +354,160 @@ func setObserved(obj *v1alpha1.Object, observed *unstructured.Unstructured) erro
...
+			c.logger.Debug("Managed resource was deleted but external resource is undeletable.")
+			isNotFound = true
+		}
+	}
+
+	if meta.WasDeleted(obj) && isNotFound {
+		// If the external resource is not found, we should remove the finalizer
+		// that was previously added to the references of this Object, since the
+		// Object, as the dependant resource, was deleted.
+		c.removeReferenceFinalizers(ctx, obj)
```
turkenh, line 472: Better to move this to `Delete()` which would simplify this method with early returns.

internal/controller/object/object.go
```diff
@@ -340,3 +354,160 @@ func setObserved(obj *v1alpha1.Object, observed *unstructured.Unstructured) erro
...
+		c.logger.Debug("Up to date!")
+
+		// We usually set condition available after create or update is invoked.
+		// However, it will never happen when policy ObserveDelete or Observe is
+		// specified since it does not allow us to create or update the external
+		// resource. In such a case, let's set condition available explicitly if
+		// the resource has been up to date.
+		if policy == v1alpha1.ObserveDelete || policy == v1alpha1.Observe {
+			obj.Status.SetConditions(xpv1.Available())
+		}
```
turkenh, line 501: May be we should always set here, for any policy. And remove the ones under Create/Update?

turkenh (commented):
Thanks @morningspace, added couple of comments mostly around styling and nitpick.

I just have some questions for the client to use reference resolution and also a concern with ignoring errors.

apis/object/v1alpha1/types.go
```diff
@@ -18,10 +18,61 @@ package v1alpha1
...
+	// Namespace of the referenced object.
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+	// FieldPath is the path of the field on the resource whose value is to be
+	// used as input.
+	// +optional
+	FieldPath *string `json:"fieldPath,omitempty"`
```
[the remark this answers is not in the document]
morningspace: This just provides the flexibility as in some cases people might just need the dependency declaration to ensure the deployment sequence. Otherwise, people have to patch something although it might not really needed in such a case.

This is a bit similar to the idea of ["SyncWaves"](https://argo-cd.readthedocs.io/en/stable/user-guide/sync-waves/) from ArgoCD where it uses annotation to declare the dependency (no patching), but ArgoCD can only control those resources that are stored in git repository. K8s provider doesn't have this limitation.

apis/object/v1alpha1/types.go
morningspace: I found that I still need `runtime.DefaultUnstructuredConverter.FromUnstructured` at last, as I need to "dump" the modified contents from `paved` to `to`, otherwise the changes will be lost.

apis/object/v1alpha1/types.go
morningspace: Totally agree! We can do it iteratively.

internal/controller/object/object.go
morningspace: You are right! My original intention was to support both Object and arbitrary K8s CR. For remote CR, it will be captured by Object from local. So, yes, all should happen at local. I've updated the code to reflect this. It actually simplifies the code.

internal/controller/object/object.go
morningspace: Nice suggestion! Just a little variant: I replaced `:` by `-`, as I found `:` is not allowed as part of finalizer name according to the spec.

internal/controller/object/object.go
morningspace: That makes perfect sense to me. Code has been updated to reflect this.

internal/controller/object/object.go
morningspace: > I don't think it is a good idea to ignore error and continue, we should return it.

I agree.

> I would already suggest to call removeReferenceFinalizers() in Delete method instead of isNotFound()

The reason I did not put in `Delete` is because, I found the reconciler will re-queue after `Delete`, then trigger another round of `Observe` against the same Object. If we remove finalizer added to referenced resource in `Delete`, it will lead to referenced resource deleted right after that. Next time when `Observe` gets called, it will always fail at reference resolving since referenced resource has gone. So, it's too early to do so in `Delete` and we should defer it as much as possible.

Actually, I found a much better place to do such thing is where after the Object finalizer is removed which is the last good time we can hook our logic. With that, the Object will be successfully deleted, so we can safely remove the finalizers added by this Object to its dependencies.

The code has been updated to reflect this change. I used a custom `Finalizer` to replace the default implementation. Both `AddFinalizer` and `RemoveFinalizer` are covered. Now the code is much better readable.

internal/controller/object/object.go
morningspace: Code has been moved to `RemoveFinalizer()`.

apis/object/v1alpha1/types.go
morningspace: Didn't quite understand this. Do you mean to mark APIVersion and Kind as optional?

morningspace:
Code has been updated according to the review comments. Now, all newly added code has been fully covered by unit tests. :-) @turkenh

turkenh (commented):
Looking pretty good @morningspace, I think we are very close to get this merged.

I did some testing this time using the manifest I used in some of my comments and working well except minor issues I've commented.

Could you also include some examples under `examples/object` and fill the `How has this code been tested` section in PR description :)

Thanks again also for your patience during this review process!

apis/object/v1alpha1/types.go
turkenh: Indeed defaulting to provider-kubernetes `Object` makes more sense since provider-kubernetes only have permission to that type by default. For accessing other resources (on local), users would need to configure RBAC.

haarchri:
@morningspace can you check latest comments ? We really want to use this feature ;)

morningspace:
Hey @haarchri, also @turkenh, sorry for my slowness! Just want to let you know that I was stuck into some project work last week, and just resume this work yesterday. I plan to finish the code change this week that will address the last couple of review comments made by @turkenh.

apis/object/v1alpha1/types.go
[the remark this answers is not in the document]
morningspace: I like the idea of using `dependsOn` and `patchesFrom`, code has been updated to reflect this.

apis/object/v1alpha1/types.go
morningspace: Agree, let's default to provider-kubernetes `Object`.

internal/controller/object/object.go
```diff
@@ -340,3 +357,177 @@ func setObserved(obj *v1alpha1.Object, observed *unstructured.Unstructured) erro
...
+func (f *objFinalizer) AddFinalizer(ctx context.Context, res resource.Object) error {
+	obj, ok := res.(*v1alpha1.Object)
+	if !ok {
+		return errors.New(errNotKubernetesObject)
+	}
+
+	if meta.FinalizerExists(obj, objFinalizerName) {
+		return nil
+	}
+	meta.AddFinalizer(obj, objFinalizerName)
```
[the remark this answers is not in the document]
morningspace: This is because, when we call `WithFinalizer` in `Setup()` to register the custom behavior on how to add and remove finalizers to and from the managed resource:

```go
managed.WithFinalizer(&objFinalizer{client: mgr.GetClient()})
```

It will replace the default behavior defined in crossplane-runtime (pkg/resource/api.go):

```go
// AddFinalizer to the supplied Managed resource.
func (a *APIFinalizer) AddFinalizer(ctx context.Context, obj Object) error {
	if meta.FinalizerExists(obj, a.finalizer) {
		return nil
	}
	meta.AddFinalizer(obj, a.finalizer)
	return errors.Wrap(a.client.Update(ctx, obj), errUpdateObject)
}
```

I know it's not ideal, but that appears to be the best option I can see :-) Any suggestion?

apis/object/v1alpha1/types.go
```diff
@@ -74,3 +138,51 @@ type ObjectList struct {
...
+// patchFieldValueToObject, given a path, value and "to" object, will
+// apply the value to the "to" object at the given path, returning
+// any errors as they occur.
+func patchFieldValueToObject(path string, value interface{}, to runtime.Object) error {
+	paved, err := fieldpath.PaveObject(to)
+	if err != nil {
+		return err
+	}
+
+	err = paved.SetValue(path, value)
```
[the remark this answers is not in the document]
morningspace: ok, fixed.

internal/controller/object/object.go
```diff
@@ -340,3 +357,177 @@ func setObserved(obj *v1alpha1.Object, observed *unstructured.Unstructured) erro
...
+	err := f.client.Update(ctx, obj)
+	if err != nil {
+		return errors.Wrap(err, errAddFinalizer)
+	}
+
+	// Add finalizer to referenced resources if not exists
+	err = f.handleRefFinalizer(ctx, obj, func(
+		ctx context.Context, res *unstructured.Unstructured, finalizer string) error {
+		if !meta.FinalizerExists(res, finalizer) {
+			meta.AddFinalizer(res, finalizer)
```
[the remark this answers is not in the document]
morningspace: That makes sense. Issue created #28.

morningspace:
Alright, I think ultimately I've made all changes needed per latest comments. @turkenh Could you please take another round of review, and see if anything missed :-)

turkenh (approved):
Looking great, thanks for incorporating all the feedback @morningspace 👍 

My only suggestion would be to consider some sort of documentation, especially how to configure RBAC if other resources need to be referenced. Your design doc is already great for anyone that wants to understand how things are working (which also needs some updates due to naming changes during implementation), but a simpler document or maybe a small section in Readme could be helpful. But this is definitely not blocking, please let me know if you want to handle this separately so that we can merge this!

cdenneen:
Instead of the patchesFrom being another object could it be from a patchSet? I’m trying to patch aws-auth which mapRoles is a Literal Scalar Block

morningspace:
@cdenneen IIUC, you are referring to the `patchSet` in `Composition` right? So, we are kind of copying stuff from `Composition` to provider level. I'd defer it till we see there's no other better option. Can you open a separate issue on this and share a concrete example w/ us? I'd like to better understand the problem and see if there's better option. Thanks.

PR #22, opened 2022-01-13
Add canonical categories to ProviderConfig

hikhvar:

Add categories to the ProviderConfig to be in line with other providers like the helm provider:
https://github.com/crossplane-contrib/provider-helm/blob/master/apis/v1beta1/types.go#L76


turkenh (approved):
Thanks!

PR #47, opened 2022-07-16
Update to go 1.17 and latest tools, add max-reconcile-rate

### LINKED ISSUE #32: Scalability: support --max-reconcile-rate arg

From https://crossplane.slack.com/archives/CEG3T90A1/p1647335790153809?thread_ts=1647335364.461669&cid=CEG3T90A1:

I'm observing that a single XRC with a composition provisionning 4 kubernetes provider Objects is taking ~30-60s to become ready on a fresh empty K8S cluster. When there is an set of existing 300 XRCs already reconciled, then a new XRC reconciles in 30 minutes on my cluster.
I'm running latest versions (crossplane 1.6.4 with helm provider  0.10 and kubernes provider 0.10.0) with the --max-reconcile-rate=100 arg in the helm chart args.

From https://github.com/crossplane/crossplane/issues/2595#issuecomment-983166375, the helm and kubernetes providers don't yet support the max reconcile rate option. 

Support for the `--max-reconcile-rate` would enable concurrent parallel processing of kubernetes provider Object MRs

bobh66:

Updated provider to go 1.17 and added the --max-reconcile-rate argument to allow for concurrent processing of up to 10 reconciliations by default.   I used the latest provider-template as a template for these changes.

Fixes #32 

I have:

The code has been running in our dev/test environment with no issues.


cmd/provider/main.go
```diff
@@ -19,29 +19,36 @@ package main
...
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
 	"github.com/crossplane/crossplane-runtime/pkg/ratelimiter"
 
 	"github.com/crossplane-contrib/provider-kubernetes/apis"
-	"github.com/crossplane-contrib/provider-kubernetes/internal/controller"
+	template "github.com/crossplane-contrib/provider-kubernetes/internal/controller"
```
turkenh, line 37:
```suggestion
	object "github.com/crossplane-contrib/provider-kubernetes/internal/controller"
```

cmd/provider/main.go
```diff
@@ -50,20 +57,43 @@ func main() {
...
+
+		// controller-runtime uses both ConfigMaps and Leases for leader
+		// election by default. Leases expire after 15 seconds, with a
+		// 10 second renewal deadline. We've observed leader loss due to
+		// renewal deadlines being exceeded when under high load - i.e.
+		// hundreds of reconciles per second and ~200rps to the API
+		// server. Switching to Leases only and longer leases appears to
+		// alleviate this.
+		LeaderElection:             *leaderElection,
+		LeaderElectionID:           "crossplane-leader-election-provider-template",
```
turkenh, line 74:
```suggestion
		LeaderElectionID:           "crossplane-leader-election-provider-kubernetes",
```

cmd/provider/main.go
```diff
@@ -50,20 +57,43 @@ func main() {
...
-	kingpin.FatalIfError(controller.Setup(mgr, log, rl, *pollInterval), "Cannot setup Template controllers")
+	o := controller.Options{
+		Logger:                  log,
+		MaxConcurrentReconciles: *maxReconcileRate,
+		PollInterval:            *pollInterval,
+		GlobalRateLimiter:       ratelimiter.NewGlobal(*maxReconcileRate),
+		Features:                &feature.Flags{},
+	}
+
+	kingpin.FatalIfError(template.Setup(mgr, o), "Cannot setup Template controllers")
```
turkenh, line 90:
```suggestion
	kingpin.FatalIfError(object.Setup(mgr, o), "Cannot setup controllers")
```

cmd/provider/main.go
```diff
@@ -19,29 +19,36 @@ package main
...
+		debug            = app.Flag("debug", "Run with debug logging.").Short('d').Bool()
+		syncInterval     = app.Flag("sync", "Controller manager sync period such as 300ms, 1.5h, or 2h45m").Short('s').Default("1h").Duration()
+		pollInterval     = app.Flag("poll", "Poll interval controls how often an individual resource should be checked for drift.").Default("1m").Duration()
+		leaderElection   = app.Flag("leader-election", "Use leader election for the controller manager.").Short('l').Default("false").Envar("LEADER_ELECTION").Bool()
+		maxReconcileRate = app.Flag("max-reconcile-rate", "The number of concurrent reconciliations that may be running at one time.").Default("10").Int()
 	)
 	kingpin.MustParse(app.Parse(os.Args[1:]))
 
-	zl := zap.New(zap.UseDevMode(*debug))
+	zl := zap.New(zap.UseDevMode(*debug), UseISO8601())
```
turkenh, line 51: What is the default value of this? 
What is the motivation behind changing the default here?

turkenh (commented):
Thanks @bobh66, left couple of minor comments!

cmd/provider/main.go
bobh66: The default for this changed between crossplane-runtime 0.14 and 0.15, I think as a result of the upgrade to zap or zapr.   The default used to be ISO8601 format, and it changed to epoch time which is probably more machine readable (though I would think a machine could consume ISO8601 just as easily) but a lot less human readable, and it makes debugging harder when trying to track down specific times.   Another option would be to only set it to ISO8601 when using debug mode if there is a reason for it to be epoch by default.

cmd/provider/main.go
bobh66: Fixed in latest patchset

cmd/provider/main.go
bobh66: fixed in latest patchset

cmd/provider/main.go
bobh66: fixed in latest patchset

turkenh (approved):

bobh66:
@turkenh - this seems to have broken the main branch because the CRDs generated by this package are missing the apiVersion field.  I see the same thing with the same changes in provider-terraform.   Any idea what could be missing/incorrect that would cause controller-gen to not include the apiVersion?

bobh66:
CRD issue Resolved by #49

PR #49, opened 2022-07-29
Remove clean.crds from Makefile - fixes #48

### LINKED ISSUE #48: Updates to latest tools broke CRD generation
### What happened?
The changes in #47 broke the main branch because the CRDs were generated with missing apiVersion attributes

This was caused by the Makefile clean.crds target being called after the generate target because it was configured as generate.done: clean,crds

Since it works in the provider-terraform project with just generate: clean.crds, which means that the CRDs are corrupted and then overwritten, I think the clean.crds target can be removed from the Makefile since whatever it was there for doesn't seem to be a problem anymore.

### How can we reproduce it?
Run make generate on the branch from #47 and observe that the generated CRDs are missing the first two lines.

bobh66:

Removed the clean.crds target from the Makefile which was removing the first two lines of the generated CRDs

Fixes #48 

I have:

Verified that the generated CRDs are now valid.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,14 +70,6 @@ cobertura:
 		grep -v zz_generated.deepcopy | \
 		$(GOCOVER_COBERTURA) > $(GO_TEST_OUTPUT)/cobertura-coverage.xml
 
-crds.clean:
-	@$(INFO) cleaning generated CRDs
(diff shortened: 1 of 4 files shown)


turkenh:
Ah sorry, I missed this 🤦 
Closing in favor of #50

PR #63, opened 2022-09-14
GKE Default Credentials and Temporary Access Tokens

### LINKED ISSUE #36: Use GKE Workload Identity for ProviderConfig Identity
We currently use GKE workload identity for creating resources with the gcp-provider. We would like to also use workload identity to authenticate to new GKE clusters that are created in order to use the kubernetes-provider (and helm-provider) with the identity GoogleApplicationCredentials type. I would assume that the identity source would be InjectedIdentity, but that seems to not be supported for the ProviderConfig identity.

Allow InjectedIdentity identity source to be used with GoogleApplicationCredentials identity type. When InjectedIdentity identity source is used perform an authentication flow using the Google Application Default Credentials that is provided by the service account's workload identity. This would also be required in the helm-provider as well.

If there are any other alternatives to using workload identity from the crossplane cluster to authenticate to a new GKE cluster, please let me know. We are not allowed to use service account keys in our GCP org and need to use workload identity for service account authentication.

bradkwadsworth-mw:
Signed-off-by: Brad Wadsworth <brad.wadsworth@mavenwave.com>

This change will allow InjectedIdentity via GKE workload identity to be used in order to authenticate to a GKE cluster. Moreover, in addition to service account keys being used as secrets, temporary access tokens may also be used. The access tokens may be created from a CronJob that regenerates the token every hour. 
Fixes #36

I have:

Tested in a GKE cluster in order to verify workload identity.

### DIFF
--- a/internal/clients/gke/gke.go
+++ b/internal/clients/gke/gke.go
@@ -49,6 +49,9 @@ func WrapRESTConfig(ctx context.Context, rc *rest.Config, credentials []byte, sc
 			t := oauth2.Token{
 				AccessToken: string(credentials),
 			}
+			if ok := t.Valid(); !ok {
+				return errors.New("Access token invalid")
+			}
 			ts = oauth2.StaticTokenSource(&t)
 		}
 	} else {
--- a/internal/clients/gke/gke.go
+++ b/internal/clients/gke/gke.go
@@ -16,6 +16,7 @@ package gke
 
 import (
 	"context"
+	"encoding/json"
 	"net/http"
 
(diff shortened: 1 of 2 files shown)


turkenh:
@bradkwadsworth-mw thanks a lot for your contributions 🙌  
Looks like this is equivalent to https://github.com/crossplane-contrib/provider-helm/pull/136. Could you update this PR with the latest state after reviews there?

turkenh:
Closing in favor of #180

PR #76, opened 2022-12-07
Fix removing finalizers when referenced resources are gone

eplightning:
Signed-off-by: Bartosz Slawianowski <bartosz.slawianowski@natzka.com>

Recently we encountered issue where sometimes a referenced object (Secret holding connection details) gets removed before the `kubernetes-provider` has a chance to remove its finalizer. This results in `Object` not being able to finish deletion process.

This PR makes it so referenced objects that no longer exist are skipped when removing finalizers. 

Let me know if you think that's not a correct approach.

I have:

Tested on local K8s cluster by deleting `Object` after manually deleting referenced resource.

### DIFF
--- a/internal/controller/object/object.go
+++ b/internal/controller/object/object.go
@@ -466,7 +466,7 @@ type objFinalizer struct {
 
 type refFinalizerFn func(context.Context, *unstructured.Unstructured, string) error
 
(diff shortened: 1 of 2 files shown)


morningspace (approved):
LGTM Thanks @eplightning

PR #99, opened 2023-02-14
Allow Object to define a set of connectionDetails

### LINKED ISSUE #15: Allows provider kubernetes to respect connection secret settings

Provider kubernetes is designed to provision and manage Kubernetes objects on (remote) Kubernetes clusters. For a normal Kubernetes object such as a ConfigMap or whatever else, it does not have to define a connection secret, since it does not make sense to use a connection string in order to connect to these Kubernetes objects. For the kubeconfig used to connect to the cluster that these Kubernetes objects reside in, it is handled separately in `ProviderConfig` using `spec. credentials`.

However, if the Kubernetes object managed by the provider is something driven by some Kubernetes controller at backend which has more complicated scenario, such as to drive a provisioning of a Kafka instance on the target cluster, then to connect to what is actually provisioned by the Kubernetes object may need some connection string, e.g. the connection string to access the Kafka instance.

Right now, the provider does not write any connection secret by itself. So, if I specify `spec.writeConnectionSecretToRef`, a connection secret with empty data will be created, which is not expected.

To allow provider kubernetes to respect the connection settings by creating the corresponding secret would help in this case. Not like other providers, the provider kubernetes itself may not know how to build such connection string, but in case the whatever backend that drives the actual workload provisioning knows that and create the secret, then the provider may have a way that can allows user to specify where to retrieve this string, e.g.: via an optional field called `spec.readConnectionSecretToRef`, then expose it to `Object` resource using `spec.writeConnectionSecretToRef`.

st3v:

This PR introduces a new property under `Object.spec`, named `connectionDetails`. This new property is optional and can be used to specify individual entries for a connection secret for the given resource.  This essentially mirrors what we already have for provider-helm (i.e. `Release.spec.connectionDetails`). As such, the changes in this PR borrow heavily from https://github.com/crossplane-contrib/provider-helm/pull/73. 

The main difference between the implementation in `provider-helm` and this, is that here we do not require the resources referred to under `connectionDetails` to have a certain set of annotations or labels. While Helm makes sure that all resources belonging to a given release have a common set of annotations, we cannot assume the same here. This is because we're dealing with all sorts of external resources here managed by arbitrary controllers that do not share a common practice when it comes to putting a well-known set of annotations or labels on the child resources they control.

This PR consists of two commits. The first one adds the new property and reconciliation logic, as well as a set of corresponding test cases and additions to the composition example. The second commit updates pre-existing parts of the example to make use of the recently introduced `Release.spec.connectionDetails[*].skipPartOfReleaseCheck` in provider-helm. This one is really just housekeeping and somewhat orthogonal to the main purpose of this PR. I'd be happy to remove it from here and submit it in a separate PR if you think that would be better. Just let me know.

Fixes #15 

I have:

* added new unit tests
* ran `make test`
* built the provider and installed it on my local cluster
* successfully verified a number of manual e2e scenarios in my local cluster


st3v:
fyi @turkenh

apis/object/v1alpha1/types.go
```diff
@@ -19,6 +19,7 @@ package v1alpha1
 import (
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
```
turkenh, line 24:
```suggestion
	v1 "k8s.io/api/core/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/apimachinery/pkg/runtime"
	
	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
```

nit: to group xp imports separately.

examples/in-composition/composition.yaml
```diff
@@ -161,19 +161,28 @@ spec:
...
               toConnectionSecretKey: argocd-ip
-             # unfortunately following is not possible today due to this limitation in provider-helm:
-             #   https://github.com/crossplane-contrib/provider-helm/blob/1d3a1c8c13a90e36c26191c70599c2e112fa10c6/pkg/controller/release/observe.go#L136
-#            - apiVersion: v1
-#              kind: Secret
-#              name: argocd-initial-admin-secret
-#              namespace: argocd
-#              fieldPath: data.password
-#              toConnectionSecretKey: argocd-password
+            - apiVersion: v1
```
turkenh, line 164: Thanks for taking care of this 🙌

turkenh (approved):
Thank you for your contribution @st3v 👍 

Please also consider adding an example, or updating an existing one similar to [here](https://github.com/crossplane-contrib/provider-helm/blob/b1f3bfc9bd0d3fe936191425a6b6f3ca2ece056b/examples/sample/release.yaml#L38). This would make the feature more discoverable.

st3v:
@turkenh You mean adding another example in addition to the arguably contrived one in my original commit?https://github.com/crossplane-contrib/provider-kubernetes/blob/bfaf1502776fb1681a7f15569ec73e68490fc61c/examples/in-composition/composition.yaml#L217

turkenh:
@turkenh You mean adding another example in addition to the arguably contrived one in my original commit?

Yeah, to make it more discoverable, we can create a directory under [example/object](https://github.com/crossplane-contrib/provider-kubernetes/tree/main/examples/object) named **connection-details** for example and put an example there.

schlakob:
Great feature! Looking forward using it! Are there any plans when this will be released?

turkenh:
@schlakob just released https://github.com/crossplane-contrib/provider-kubernetes/releases/tag/v0.8.0 🚀

PR #100, opened 2023-02-17
upgrade : go1.20, golangci to  v1.51.2

fahedouch:

- `go.mod` still pointing to `go1.19` to highlight the minimum supported version of Go.
- upgrade golangci to `v1.51.1`

-------------------
_edit_
upgrade golangci to `v1.51.2`

I have:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,8 +10,8 @@ on:
 
 env:
   # Common versions
-  GO_VERSION: '1.19.5'
-  GOLANGCI_VERSION: 'v1.50.1'
+  GO_VERSION: '1.20.2'
+  GOLANGCI_VERSION: 'v1.51.2'
   DOCKER_BUILDX_VERSION: 'v0.8.2'
 
   # Common users. We can't run a step 'if secrets.AWS_USR != ""' but we can run
--- a/README.md
+++ b/README.md
@@ -30,6 +30,8 @@ spec:
 
 ## Developing locally
 
+See the header of [`go.mod`](./go.mod) for the minimum supported version of Go.
+
 Start a local development environment with Kind where `crossplane` is installed:
 
 ```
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -10,8 +10,8 @@ on:
 
 env:
   # Common versions
-  GO_VERSION: '1.19.5'
-  GOLANGCI_VERSION: 'v1.50.1'
+  GO_VERSION: '1.20.x'
```
turkenh, line 13: Is there a feature that we want to get with this new version? Otherwise, I am wondering if we should wait a bit to get  it stabilized for a while?

README.md
```diff
@@ -30,6 +30,8 @@ spec:
 
 ## Developing locally
 
+See the header of [`go.mod`](./go.mod) for the minimum supported version of Go.
```
turkenh, line 33: I would instead prefer to be consistent with the `go.mod`, would it fail with go `v1.19` if we update `go.mod` too?

.github/workflows/ci.yml
turkenh: Also, having a fixed version would produce more consistent builds, i.e. `1.20.1` instead of `1.20.x`

fahedouch: > Is there a feature that we want to get with this new version? 

Nope.

> Otherwise, I am wondering if we should wait a bit to get it stabilized for a while?

Sure

README.md
fahedouch: the go version inside `go.mod` is intended to be the minimum supported version ( [see](https://go.dev/doc/modules/gomod-ref) ).

So as the minimum supported version `go mod tidy` will fail it the version is not matched ( < `supported version`) 

```
faheddorgaa@MBP-de-fahed: go version
go version go1.19 darwin/amd64

faheddorgaa@MBP-de-fahed:$ go mod tidy
go: go.mod file indicates go 1.20, but maximum version supported by tidy is 1.19
```

.github/workflows/ci.yml
fahedouch: @turkenh fixed.

turkenh (approved):
Thank you @fahedouch 🙌 

Could you also set the linter version in the makefile like [here](https://github.com/crossplane/crossplane-runtime/blob/cf3c7a09628aa81df570738af885a4b930343905/Makefile#L37) ?

fahedouch:
> 

Hi @turkenh, done. go upgraded to `1.20.2` too

PR #116, opened 2023-05-08
Use lazy rest mapper for performance and less audit

chlunde:

provider-kubernetes initialized a new kubernetes client for each
reconcile. The REST mapper in controller-runtime used to fetch
information about every CRD in the cluster.

controller-runtime introduced a lazy restmapper which means we don't
have to introduce any complex caching to get a significant performance
boost in provider-kubernetes:

https://github.com/kubernetes-sigs/controller-runtime/pull/2116

This seems to become the default in the next release:

https://github.com/kubernetes-sigs/controller-runtime/pull/2296

But this is so significant that we want to update now:

* CPU reduced from constant throttling at 0.4 cores to 0.04 cores

* CloudWatch / EKS audit log costs reduced significantly (55% for our
  cluster, with a lot of provider-kubernetes resources)

I have:

Run in cluster

### DIFF
--- a/internal/clients/client.go
+++ b/internal/clients/client.go
@@ -19,6 +19,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/clientcmd/api"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
 // NewRESTConfig returns a rest config given a secret with connection information.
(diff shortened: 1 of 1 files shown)


chlunde:
@morningspace could you take a look at this? Thanks!

internal/clients/client.go
```diff
@@ -33,7 +34,14 @@ func NewRESTConfig(kubeconfig []byte) (*rest.Config, error) {
 // NewKubeClient returns a kubernetes client given a secret with connection
 // information.
 func NewKubeClient(config *rest.Config) (client.Client, error) {
-	kc, err := client.New(config, client.Options{})
+	// in 0.15 this will be default
```
turkenh, line 37: Just merged a bump PR which updates controller runtime to [v0.15.1](https://github.com/crossplane-contrib/provider-kubernetes/blob/88915189d3837feb6ac826ef65a4061f843330a7/go.mod#L17), does this mean we no longer need the changes here?

phisco: yes: https://github.com/kubernetes-sigs/controller-runtime/blob/40203bfeac472914f36ea35bf386326814f6f82e/pkg/client/client.go#L150

PR #119, opened 2023-05-23
feat: deriving readiness from object

### LINKED ISSUE #45: Health checks for resources
For our case, it's critical to be able to perform health assessment of resources created from Objects. Specifically, for custom resources compatible with [kstatus](https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus#the-ready-condition), which essentially means "wait till the resource's status has `Ready = True` condition".

I would assume the correct implementation means the following:

- there's some way to enable health check for arbitrary Object
- if enabled, Object doesn't become ready until underlying resource reports Ready status
- Objects with dependencies (ones that have references[].dependsOn) can honor status of their dependency, regardless of whether it's another Object or an arbitrary k8s resource

I apologize if it's already possible, I didn't realize how to do it.

### LINKED ISSUE #68: Align Object readiness to Manifest readiness

One of the actual limitations of Crossplane is that existing Kubernetes operator (i.e. check operatorhub.io) cannot be integrated within a Composition because the provider-kubernetes - in particular, the Kind Object - is not aware of the Readiness of the applied manifest. The actual implementation of Kind Object has status Ready when the Kubernetes manifest is applied.

The Ready status of the Kind Object should be aligned with the status of the applied Manifest. This would expand the compatibility of the Composition with operators that are not implement with the Crossplane runtime.

phisco:

Fixes #68
Fixes #45

Introduces a new field in the Object spec allowing to specify whether the Ready condition's status of the Object should reflect the one of the Managed Resource.

The reason is not propagated, it's only mapped to "Available"/"Unavailable" as per https://github.com/crossplane/crossplane/blob/master/contributing/guide-provider-development.md?plain=1#L457, but if we agree we could forward that too if present.

I have:

locally deploying the provider an crossplane, applying the following manifest:
```yaml
kubectl apply -f - <<EOF
kind: Object
metadata:
  name: sample-pod
spec:
  readiness:
    policy: DeriveFromObject # default is SuccessfulCreate, which is the current behaviour
  forProvider:
    manifest:
      apiVersion: v1
      kind: Pod
      metadata:
        namespace: default
        labels:
          run: pod
      spec:
        initContainers:
        - image: busybox
          name: sleep
          command: ['sh', '-c', '--']
          args: ['sleep 30 && echo "done"']            #30s of init, not yet ready
        containers:
        - image: busybox
          name: sleepboom
          command: ['sh', '-c', '--']
          args: ['sleep 90; exit 1']                              #90s ready and then error out
        restartPolicy: Never                                      #never recovered, so the readiness of the Object should be set to false.
  providerConfigRef:
    name: kubernetes-provider
EOF
```

Properly results in the readiness of the Object being set both ways:
```shell
$ kubectl get pods -w
NAME         READY   STATUS    RESTARTS   AGE
sample-pod   0/1     Pending   0          0s
sample-pod   0/1     Pending   0          0s
sample-pod   0/1     Init:0/1   0          0s
sample-pod   0/1     Init:0/1   0          2s
sample-pod   0/1     PodInitializing   0          32s    # init container returns after 30s
sample-pod   1/1     Running           0          34s      # pod is ready immediately after that around 30s+
sample-pod   0/1     Error             0          2m4s      # pod crashes and is never restarted around ~2m(=30s+90s)
sample-pod   0/1     Error             0          2m5s
```
While on the Object, given that the default PollInterval is set to `60s`, the readiness will be propagated with some delay:
```shell
k get objects.kubernetes.crossplane.io -w                                                              1 ↵
NAME         KIND   PROVIDERCONFIG        SYNCED   READY   AGE
sample-pod   Pod    kubernetes-provider                    0s
sample-pod   Pod    kubernetes-provider                    0s
sample-pod   Pod    kubernetes-provider                    0s
sample-pod   Pod    kubernetes-provider                    0s
sample-pod   Pod    kubernetes-provider                    0s
sample-pod   Pod    kubernetes-provider   True     False   0s
sample-pod   Pod    kubernetes-provider   True     False   0s
sample-pod   Pod    kubernetes-provider   True     True    61s       # finally notices the pod is ready ~60s(=pollInterval)
sample-pod   Pod    kubernetes-provider   True     False   3m1s    # notices the pod is not ready anymore ~3m(=30s+60s, rounded to the next pollInterval)
```

While without setting `readiness.policy` to `SuccessfulCreate` (default and current behaviour), the Object immediately becomes Ready as soon as the Pod is successfully applied, even if it crashes straight away.


apis/object/v1alpha1/types.go
```diff
@@ -123,6 +122,23 @@ type ObjectSpec struct {
...
+}
+
+// Readiness defines how the object's readiness condition should be computed,
+// if not specified it will be considered ready as soon as the underlying external
+// resource is considered up-to-date.
+type Readiness struct {
+	// FromObject configures whether the Object's readiness should be derived from
+	// the readiness of the underlying external resource.
+	// +optional
+	FromObject *bool `json:"fromObject,omitempty"`
```
negz, line 135: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types

Not a hard and fast rule, but it's generally a good idea to avoid boolean fields.

> Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options. Plan for future expansions by describing the policy options explicitly as a string type alias (e.g. TerminationMessagePolicy).

I could see this being another enum like `readiness.policy` or `readiness.source` with two options (for now):

1. Successful creation of object (I think this is the default today, right? I didn't check).
1. Derive from readiness of object. (Would this also need optional configuration - i.e. what indicates the object is ready?)

phisco: Our idea was to leave it to a simple Boolean now, but put it in a dedicated stanza that would allow us to add a list of readinessChecks if required down the line:

```yaml
readiness:
  fromObject: false
  checks:
  - type: MatchCondition
    matchCondition: ...
```

But we could go with:

```yaml
readiness:
  type: FromChecks #FromObject
  checks:
  - type: MatchCondition
    matchCondition: ...
```
Which would mean now we only implement:

```yaml
readiness:
  type: FromObject
```

turkenh: Throwing out another possibility.

```
readiness:
  policy: SuccessfullCreate (default) # DeriveFromObject, PassingChecks
```

phisco: which would be self-documenting w.r.t. the current behaviour too, love it!

jbw976:
@phisco is this a scoped implementation just here in provider-kubernetes to address https://github.com/crossplane/crossplane/issues/4091?

apis/object/v1alpha1/types.go
phisco: implemented and updated the PR description accordingly, kept it in a separate commit, but I'd squash everything before merging.

phisco:
@jbw976 -
> phisco is this a scoped implementation just here in provider-kubernetes to address [crossplane/crossplane#4091](https://github.com/crossplane/crossplane/issues/4091)?

The original issues, #45 and #68, were only asking for a way to have the external resource's Ready condition propagated to the Object. We initially thought we could generalise it and add a readinessCheck type to Crossplane allowing to inspect conditions at an arbitrary path, hence https://github.com/crossplane/crossplane/issues/4091.

But then @negz pointed out (https://github.com/crossplane/crossplane/issues/4091#issuecomment-1558139291) that actually the need to propagate readiness from arbitrary paths to conditions was specific to `provider-kubernetes`, so we agreed it still made sense to add a new readiness check type `MatchCondition` to Crossplane which would both allow customising the Condition to match, but also be self-documenting w.r.t. the current behaviour.

So, to sum it up:
- this change on `provider-kubernetes` alone is enough to solve #45 and #68
- [crossplane/crossplane#4091] and the PR I'm working on is just a nice-to-have on the Crossplane side to improve the user experience

internal/controller/object/object.go
```diff
@@ -358,6 +358,16 @@ func setObserved(obj *v1alpha1.Object, observed *unstructured.Unstructured) erro
 	if obj.Status.AtProvider.Manifest.Raw, err = observed.MarshalJSON(); err != nil {
 		return errors.Wrap(err, errFailedToMarshalExisting)
 	}
+
+	if obj.Spec.Readiness.Policy == v1alpha1.ReadinessPolicyDeriveFromObject {
+		conditioned := xpv1.ConditionedStatus{}
+		err := fieldpath.Pave(observed.Object).GetValueInto("status", &conditioned)
```
turkenh, line 364: Could logging the error be helpful for debugging purposes? I see a couple of cases that might cause an error here, and seeing it in the logs or events would be beneficial rather than just seeing the resource as unavailable no matter what.

turkenh (approved):
Thanks @phisco 🙌

PR #118, opened 2023-05-18
Setup e2e with Local Deploy and Uptest

turkenh:
### Description of your changes

This PR configures the `e2e` target to run for each PR validating:

- Provider is packaged properly
- Provider runs successfully
- An example `Object` processed and become ready.

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

```
make e2e
```

[contribution process]: https://git.io/fj2m9

PR #120, opened 2023-05-23
docs: switch to crossplane-system namespace everywhere for consistency

phisco:

Switching all occurrences of `upbound-system` namespace back to `crossplane-namespace`.

I have:


bobh66:
> crossplane-system is not created anymore, 

Since when?  Community crossplane deploys into crossplane-system, and that should remain the default for all non-Upbound providers.

upbound-system is only used for downstream uxp and should not be referenced in anything upstream of uxp

turkenh:
> Since when? Community crossplane deploys into crossplane-system, and that should remain the default for all non-Upbound providers.

> upbound-system is only used for downstream uxp and should not be referenced in anything upstream of uxp

Agreed. 

This was changed for local development recently for integrating uptest for e2e testing: https://github.com/crossplane-contrib/provider-kubernetes/pull/118

But, I believe we should rather configure uptest to use `crossplane-system` instead.

phisco:
good point @bobh66, totally agree, my main concern was about consistency and the procedure in the README properly working. I switched everything to `crossplane-system` and tomorrow we'll fix the build sub module to properly handle CROSSPLANE_NAMESPACE 👍

turkenh (approved):
Thanks @phisco 🙌

phisco:
Just noticed the cleanup `make local.down` in the README doesn't work for me.
I'm not sure which would be the correct target to use for that though 🤔 
CC @turkenh

turkenh:
@phisco I assume it should be updated as `make controlplane.down`.

phisco:
pushed the change, it works fine now, totally missed that target, thanks @turkenh!

PR #110, opened 2023-03-31
WIP: Add DesiredStateChanged event filter

bobh66:
### Description of your changes
Added the DesiredStateChanged EventFilter to the controller to avoid unnecessary reconciliations.

I have:

- [X] Read and followed Crossplane's [contribution process].
- [X] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested
Executed this code in a development cluster doing multiple deployments.

[contribution process]: https://git.io/fj2m9

PR #122, opened 2023-06-27
cleanerr: Clean errors for stable status

chlunde:

Some error messages from the kubernetes API contain formatted pointers. This
triggers an infinite loop in provider-kubernetes, as each attempt to reconcile
updates the status with a new error message.

To avoid this loop, mask pointer references in errors on update.

A similar fix exists in provider-aws.

Without this change, we can see a lot of reconciles for one object over a short
time period:

  Warning  CannotUpdateExternalResource  3m38s (x849 over 8m39s)  managed/object.kubernetes.crossplane.io  (combined from similar events):

This might also block other resources from reconciling.

For consideration: It may be that PR https://github.com/crossplane-contrib/provider-kubernetes/pull/110 effectively fixes _most_ of the issue covered here. Events produced would still be duplicated, but the reconcile spinning would likely stop.

I have:

* Ran provider-kubernetes locally, created Job objects and modified them.
* Added new unit tests for this function.

### DIFF
--- a/internal/controller/object/cleanerr.go
+++ b/internal/controller/object/cleanerr.go
@@ -1,0 +1,37 @@
+package object
+
+import "regexp"
+
+// CleanErr masks pointers in err when printing the message.
+// errors.As and errors.Is can still find the original error.
+// If err is nil, CleanErr returns nil.
+func CleanErr(err error) error {
+	if err == nil {
+		return nil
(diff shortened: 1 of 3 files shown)


turkenh (approved):
Thanks @chlunde 🙌

PR #128, opened 2023-08-15
Bump Crossplane Runtime to the Latest

turkenh:

This PR consumes the latest Crossplane Runtime with [this PR](https://github.com/crossplane/crossplane-runtime/pull/513) and drops dependencies to multiple Hashicorp repositories.

I have:

```
make reviewable build
```


apis/object/v1alpha1/management_policy_hack.go
```diff
@@ -68,14 +68,14 @@ type ResourceSpec struct {
 	DeletionPolicy xpv1.DeletionPolicy `json:"deletionPolicy,omitempty"`
 }
 
-// GetManagementPolicy of this Object.
-func (mg *Object) GetManagementPolicy() xpv1.ManagementPolicy {
+// GetManagementPolicies of this Object.
+func (mg *Object) GetManagementPolicies() xpv1.ManagementPolicies {
```
jbw976, line 72: can you remind me if there are migration concerns for this update? e.g. do users of this provider need to do anything on their side?

jbw976 (approved):

apis/object/v1alpha1/management_policy_hack.go
turkenh: No.
We need to deprecate the `managementPolicy` in this provider in favor of `managementPolicies` in Crossplane Runtime at some point (and think about migration). However, this is not included in this PR.

PR #170, opened 2023-12-19
feat(kubelogin): add kubelogin azure ad auth

### LINKED ISSUE #105: Integrate kubelogin for Azure AD Authentication to AKS Clusters
We are provisioning Azure AKS Kubernetes Clusters using the [Azure Provider by Upbound](https://marketplace.upbound.io/providers/upbound/provider-azure/). After provisioning, we install software inside of the clusters using this Kubernetes Provider as well as the related [Helm Provider](https://github.com/crossplane-contrib/provider-helm). When using Local Accounts for authentication with the clusters API servers, this works out of the box. However, we would like to switch to authentication based on Azure AD Service Principals and Managed Identities, as this is the best practice and recommended by Azure. The [Kubernetes Go Client](k8s.io/client-go) which is used by this provider does not support this authentication method by default, so currently we can't use this provider with AKS clusters configured with Azure AD Authentication. As the same problem arises with standard kubectl usage and Terraform providers as well (see e.g. https://github.com/Azure/kubelogin/issues/114), Azure provides the [kubelogin](https://github.com/Azure/kubelogin) client-go credential plugin which can be used to retrieve the user credentials and pass it to the Kubernetes Go Client.

I propose to integrate the [official kubelogin Go package](https://pkg.go.dev/github.com/Azure/kubelogin) into this provider (as well as the Helm Provider), so that it's possible to use it with AKS clusters configured with Azure AD Authentication. I'm open to provide an implementation for this and open a PR. Also, I would be happy to hear the maintainer's thoughts about this. Are you open to such contribution? @turkenh @morningspace

haarchri:
initial work done in https://github.com/crossplane-contrib/provider-helm/pull/205 by @erhancagirici

Enables configuring Azure AD authentication via `kubelogin` integration. Introduces the type `AzureServicePrincipalCredentials` in `ProviderConfig` `.spec.identity` section.

The specified Azure Service Principal credentials provided via Secret, for authenticating to AKS cluster by obtaining a token through `kubelogin`

An example ProviderConfig using `AzurePrincipalCredentials` as identity for authenticating to AzureAD

```
kind: ProviderConfig
metadata:
  name: kubernetes-provider
spec:
  credentials:
    source: Secret
    secretRef:
      namespace: crossplane-system
      name: cluster-config
      key: kubeconfig
  identity:
    type: AzureServicePrincipalCredentials
    source: Secret
    secretRef:
      name: azure-credentials
      namespace: crossplane-system
      key: credentials.json

```

Fixes #105

I have:

test image: index.docker.io/haarchri/provider-kubernetes:v0.0.0-130.g8467c56

```
 kubectl get objects
NAME                   KIND             PROVIDERCONFIG   SYNCED   READY   AGE
test-aks-5qft5         ApplicationSet   test-aks         True     True    16m
test-aks-f6mcm-gqthg   Secret           default          True     True    14m
```

```
- apiVersion: kubernetes.crossplane.io/v1alpha1
  kind: ProviderConfig
  metadata:
    name: test-aks
  spec:
    credentials:
      secretRef:
        key: kubeconfig
        name: 3a02bacd-0d8a-40b2-8922-560ba4acf7ed-akscluster
        namespace: upbound-system
      source: Secret
    identity:
      secretRef:
        key: credentials
        name: azure-creds
        namespace: upbound-system
      source: Secret
      type: AzureServicePrincipalCredentials
```

```
- apiVersion: kubernetes.crossplane.io/v1alpha1
  kind: ProviderConfig
  metadata:
    name: default
  spec:
    credentials:
      source: InjectedIdentity
```


haarchri:
this error https://github.com/crossplane-contrib/provider-kubernetes/actions/runs/7267353656/job/19800977910?pr=170#step:11:1005 is related to: https://github.com/upbound/uptest/pull/169

internal/clients/azure/azure.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+// WrapRESTConfig configures the supplied REST config to use OAuth2 bearer
+// tokens fetched using the supplied Azure Credentials.
+func WrapRESTConfig(_ context.Context, rc *rest.Config, credentials []byte, _ ...string) error {
+	m := map[string]string{}
+	if err := json.Unmarshal(credentials, &m); err != nil {
+		return err
+	}
+
```
erhancagirici, line 30: In this implementation, we enter this codepath if the consumer set an `identity` section in provider config, then we unconditionally try to process the kubeconfig even if it can work standalone. In that case, ideally user should not be specifying an `identity` section, but when they do by mistake, this code does not properly handle it. rest See my open issue regarding the similar implementation at provider-helm: https://github.com/crossplane-contrib/provider-helm/issues/206

At this point, I think we should do nil checks for `rc.ExecProvider` and `rc.ExecProvider.Args` otherwise we get nil pointer err below. Then, for the nil check we have two options:
1. since user specified an `identity` section, we assume that they intend to configure an azure ad auth setup. Therefore, we return an error, and say that this kubeconfig is not in expected format (does not have exec provider section)
2. since `execProvider` is not present, we assume that the kubeconfig is a standalone type, and no processing should/can be done here, hence the specified `identity` section is considered superfluous and ignored. We return here with no error. If there are other problems with kubeconfig unrelated to this, let them surface while connecting.

Although not strongly opinionated, I am more inclined towards option 1, as supplying an `identity` section should be considered as an intention to use an azureAD enabled cluster, and the kubeconfig should conform to that, not vice-versa. Even if a working standalone kubeconfig was provided, we should notify user with an error and make sure of their intention, they should either fix their kubeconfig or do not specify any identity configuration.
something like:
```suggestion
if rc.ExecProvider == nil || rc.ExecProvider.Args == nil || len(execProvider.Args) < 1 {
    return errors.New("an identity configuration was specified but the provided kubeconfig does not have execProvider section")
}
```
WDYT?

erhancagirici (commented):
Left one comment, otherwise LGTM

internal/clients/azure/azure.go
turkenh: +1 for option 1.

turkenh (approved):
Thanks @haarchri!

internal/clients/azure/azure.go
haarchri: implemented

PR #190, opened 2024-02-02
feat(in-cluster): update example to use native crossplane way

haarchri:
use crossplane native way to grant a static service account the needed cluster-admin rights

Fixes #

I have:

### DIFF
--- a/examples/provider/config-in-cluster.yaml
+++ b/examples/provider/config-in-cluster.yaml
@@ -1,4 +1,4 @@
-## check provider-in-cluster.yaml for grant access
+# Check ./provider-in-cluster.yaml to see how to grant permissions to the Provider
 apiVersion: kubernetes.crossplane.io/v1alpha1
 kind: ProviderConfig
 metadata:
--- a/examples/provider/config-in-cluster.yaml
+++ b/examples/provider/config-in-cluster.yaml
@@ -1,3 +1,4 @@
+## check provider-in-cluster.yaml for grant access
 apiVersion: kubernetes.crossplane.io/v1alpha1
 kind: ProviderConfig
 metadata:
@@ -5,29 +6,3 @@ metadata:
 spec:
   credentials:
     source: InjectedIdentity
-  runtimeConfigRef:
-    apiVersion: pkg.crossplane.io/v1beta1
-    kind: DeploymentRuntimeConfig
-    name: provider-kubernetes
----
-apiVersion: pkg.crossplane.io/v1beta1
-kind: DeploymentRuntimeConfig
-metadata:
-  name: provider-kubernetes
-spec:
-  serviceAccountTemplate:
-    metadata:
-      name: provider-kubernetes
----
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
-metadata:
-  name: provider-kubernetes-cluster-admin
-subjects:
-  - kind: ServiceAccount
-    name: provider-kubernetes
-    namespace: crossplane-system
-roleRef:
-  kind: ClusterRole
-  name: cluster-admin
-  apiGroup: rbac.authorization.k8s.io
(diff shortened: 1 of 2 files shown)


examples/provider/config-in-cluster.yaml
```diff
@@ -1,12 +1,33 @@
...
 kind: ProviderConfig
 metadata:
   name: kubernetes-provider
 spec:
   credentials:
     source: InjectedIdentity
+  runtimeConfigRef:
+    apiVersion: pkg.crossplane.io/v1beta1
+    kind: DeploymentRuntimeConfig
+    name: provider-kubernetes
```
phisco, line 11: I think this needs to be on the Provider, not on the ProviderConfig

examples/provider/provider-in-cluster.yaml
```diff
@@ -0,0 +1,33 @@
+---
+apiVersion: pkg.crossplane.io/v1
+kind: Provider
+metadata:
+  name: provider-kubernetes
+spec:
+  package: xpkg.upbound.io/crossplane-contrib/provider-kubernetes:v0.10.0
```
phisco, line 7: We should definitely add renovate to this repo, in the meantime let's start from the latest version:
```suggestion
  package: xpkg.upbound.io/crossplane-contrib/provider-kubernetes:v0.11.4
```

examples/provider/config-in-cluster.yaml
```diff
@@ -1,12 +1,8 @@
-# Make sure provider-kubernetes has enough permissions to install your objects into cluster
-#
-# You can give admin permissions by running:
-# SA=$(kubectl -n crossplane-system get sa -o name | grep provider-kubernetes | sed -e 's|serviceaccount\/|crossplane-system:|g')
-# kubectl create clusterrolebinding provider-kubernetes-admin-binding --clusterrole cluster-admin --serviceaccount="${SA}"
+## check provider-in-cluster.yaml for grant access
```
phisco, line 1:
```suggestion
# Check ./provider-in-cluster.yaml to see how to grant permissions to the Provider
```

examples/provider/config-in-cluster.yaml
haarchri: Small hickup during commit from flight ...

examples/provider/provider-in-cluster.yaml
```diff
@@ -0,0 +1,33 @@
+---
+apiVersion: pkg.crossplane.io/v1
```
bobh66, line 2: Would it make more sense/be easier to read if the resources are listed in the order that they need to be created?  I know reconciliation will resolve ordering issues, but it might be more readable if it starts with `ServiceAccount` then `ClusterRoleBinding` then `DeploymentRuntimeConfig` and then `Provider`.

examples/provider/provider-in-cluster.yaml
```diff
@@ -0,0 +1,33 @@
...
+    metadata:
+      name: provider-kubernetes
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: provider-kubernetes-cluster-admin
+subjects:
+  - kind: ServiceAccount
+    name: provider-kubernetes
```
bobh66, line 28: The referenced Service Account is not created by this example - does it need to be for completeness?

haarchri: RuntimeConfig will create the ServiceAccount for you

examples/provider/provider-in-cluster.yaml
haarchri: Split the File only for ordered creation - can add it in one file with an comment

haarchri: We don't need to create the ServiceAccount - RuntimeConfig will create the ServiceAccount for you

turkenh (approved):
Thanks, looking good to me!

PR #193, opened 2024-02-07
enable opt-in to redact Secret data from Object status

ravilr:
### Description of your changes

Fixes https://github.com/crossplane-contrib/provider-kubernetes/issues/178

- Adds a new flag `--sanitize-secrets`  when set in controllerConfig/deploymentRuntimeConfig of provider-kubernetes, redacts the Secret data from Object `status.atProvider.manifest`.

- the new redact behavior is off by default so as to not break any existing compositions relying on `status.atProvider.manifest.data` for patches.  Usages meant to shuttling observed secret data to/from remote/control-plane clusters can be updated to use  `spec.references[].patchesFrom` or `spec.connectionDetails` and avoid leaking secret data to Object MR status resources.

I have:

- [X] Read and followed Crossplane's [contribution process].
- [X] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

- `make test` and `make e2e`

Updating the provider-kubernetes's deploymentRuntimeConfig with below additional arg:
```
         args:
           ......
+          - '--sanitize-secrets'
```

```
$ kubectl get provider.pkg provider-kubernetes
NAME                  INSTALLED   HEALTHY   PACKAGE                                           
provider-kubernetes   True        True      ravilr/provider-kubernetes:v0.11.6
```

will reconcile any existing and new Object resources managing Secrets to have their `status.atProvider.manifest.data` redacted:

```
apiVersion: kubernetes.crossplane.io/v1alpha2
kind: Object
metadata:
  annotations:
    crossplane.io/composition-resource-name: credential-secret
    crossplane.io/external-create-pending: "2024-02-07T07:29:20Z"
    crossplane.io/external-create-succeeded: "2024-02-07T07:29:20Z"
    crossplane.io/external-name: provider-token-nkm7s-9wlpl
  creationTimestamp: "2024-02-07T07:29:21Z"
  finalizers:
  - finalizer.managedresource.crossplane.io
  generateName: provider-token-nkm7s-
  generation: 2
  labels:
    crossplane.io/claim-name: provider-token
    crossplane.io/claim-namespace: awsuswest2-admin
    crossplane.io/composite: provider-token-nkm7s
  name: provider-token-nkm7s-9wlpl
  ownerReferences:
  - apiVersion: security.example.com/v1
    blockOwnerDeletion: true
    controller: true
    kind: XCredential
    name: provider-token-nkm7s
    uid: 1569a117-9227-4567-8db3-259e18171da5
  resourceVersion: "86940719"
[...]

PR #198, opened 2024-02-13
Add jitter to poll interval

ravilr:

- Adds a +/-10% jitter to Object's reconcile delay, helps towards spreading out requeue'ing of resources and minimize the likelihood of running into K8s client qps/burst limits.

I have:

- `make test` and `make e2e`

- below graph shows the before(blue plot) and after(green plot) effect of this change on reconciliation rate of Object resources running with poll-interval=5m and poll-jitter=30s:
Screenshot 2024-02-10 at 2 11 49 PM

### DIFF
--- a/cmd/provider/main.go
+++ b/cmd/provider/main.go
@@ -72,6 +72,14 @@ func main() {
 		ctrl.SetLogger(zl)
 	}
 
+	// configure the jitter to be the 10% of the poll interval
+	pollJitter := time.Duration(float64(*pollInterval) * 0.1)
+	log.Debug("Starting",
+		"sync-interval", syncInterval.String(),
+		"poll-interval", pollInterval.String(),
+		"poll-jitter", pollJitter.String(),
+		"max-reconcile-rate", *maxReconcileRate)
+
 	cfg, err := ctrl.GetConfig()
 	kingpin.FatalIfError(err, "Cannot get API server rest config")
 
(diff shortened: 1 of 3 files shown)


ravilr:
@lsviben @turkenh PTAL 🙏

cmd/provider/main.go
```diff
@@ -72,6 +72,14 @@ func main() {
 		ctrl.SetLogger(zl)
 	}
 
+	// configure the jitter to be the 10% of the poll interval
```
lsviben, line 75: NIT: some other providers I checked all have it set to 5%. Unless there is any reason not to, it would be cool to have it here as well. Saying that, I am not sure why 5% was chosen instead of some other number.

lsviben (approved):
Thanks @ravilr ! LGTM

cmd/provider/main.go
ravilr: sure. I picked 10% based on [provider-terraform](https://github.com/upbound/provider-terraform/pull/204) change, but without exposing it as a configurable flag option. Also, based on the observation on a cluster that 10% jitter leads to noticeable reconcile intervals spread (the graph attached in description).

if we think this needs to be made configurable through flag option, i can add that.

lsviben: If you could add it as a flag, that would be awesome. The default can be 10%

ravilr: added `--poll-jitter-percentage` flag.

ravilr:
@lsviben  just checking when is the next planned release for this provider ?  Interested in this and #193  included in a release. Thanks.

turkenh:
@ravilr doing some performance/load testing with this provider. 
I hope to have a release with some improvements from there later this week or early next week.

PR #217, opened 2024-03-28
Add `ObservedObjectCollection` API type

### LINKED ISSUE #209: Support observing child resources

It's common for folks to use this provider to create Kubernetes resources like Deployments and CronJobs. They do this by wrapping the Deployment (for example) in an Object MR.

Often these resources produce child resources. For example a CronJob will produce Jobs, and a Deployment will produce ReplicaSets and Pods. Today there's no way to observe the status of these children.

Consider a composition function that composes a CronJob, and wants expose the status of the resulting Jobs. Given that functions:

* [Must not](https://github.com/crossplane/crossplane/blob/master/contributing/specifications/functions.md#runtime-environment) access the API server directly
* Often won't be running in the same cluster as the one where provider-kubernetes created the CronJob

There's no obvious way today to load up Jobs etc. You could compose observe-only Objects, but how would you know how many Objects to create and what Job names they should observe?

One approach could be to add an ObjectCollection MR. Such an MR would be:

* Inherently observe-only. It would not support creating, updating, or deleting objects
* Would accept a GVK and a label selector, and expose all matching resources in its status

A special case version would be EventCollection, for loading events. This would be similar to an ObjectCollection but it would match on the resource the events pertained to, rather than labels.

pedjak:

Objects in the collection are defined by:
* GVK
* optional namespace
* label selector

```yaml
kind: ObservedObjectCollection
metadata:
  name: jobs
  labels:
    foo: bar
spec:
  observeObjects:
    apiVersion: batch/v1
    kind: Job
    namespace: my-ns
    selector:
      matchLabels:
          foo: bar
  providerConfigRef:
     name: kubernetes-provider
  objectTemplate:
    metadata:
      labels:
        foo: bar
```

The provider reconcile the collection by fetching objects matching the selector and creates counterpart "observe-only" Objects in the local cluster. Additional labels and annotations can be added to them by specifying `objectTemplate.metadata` on the collection resource.

```yaml
kind: Object
metadata:
  name: 00647af6-a82f-4344-98b3-4e5dcbbbfa8d
  labels:
    kubernetes.crossplane.io/owned-by-collection: jobs
    foo:bar
spec:
  managementPolicies: ["Observe"]
  forProvider:
    manifest:
      apiVersion: batch/v1
      kind: Job
      metadata:
        name: job1
        namespace: my-ns
  providerConfigRef:
    name: kubernetes-provider
```

```yaml
kind: Object
metadata:
  name: ae2d702b-e468-41fa-b427-f96ee03928aa
  labels:
    kubernetes.crossplane.io/owned-by-collection: jobs
    foo: bar
spec:
  managementPolicies: ["Observe"]
  forProvider:
    manifest:
      apiVersion: batch/v1
      kind: Job
      metadata:
        name: job2
        namespace: my-ns
  providerConfigRef:
    name: kubernetes-provider
```

* The created objects are owned by the collection resource and reconciled as usual by the provider. 
* Each created object is accompanied with `kubernetes.crossplane.io/owned-by-collection` label uniquely identifying the collection object, so that clients can use label to obtain all members of the given collection
* That label is published on collection as well, under `status.membershipLabel`

```yaml
kind: ObservedObjectCollection
metadata:
  name: jobs
spec:
  observeObjects:
    apiVersion: batch/v1
    kind: Job
    namespace: my-ns
    selector:
      matchLabels:
          foo: bar
  providerConfigRef:
     name: kubernetes-provider
status:
  membershipLabel:
    kubernetes.crossplane.io/owned-by-collection: jobs
```

```mermaid
flowchart LR;
collection["ObservedObjectCollection"]
provider{{provider-kubernetes}}
cluster((k8s cluster))
object1["Object

managementPolicies: Observe"]
object2["Object

managementPolicies: Observe"]
object3["Object

managementPolicies: Observe"]

provider-->|reconciles|collection
provider-->|list objects matching selector|cluster
provider-->|creates|object1
provider-->|creates|object2
provider-->|creates|object3
collection-->|refers|object1
collection-->|refers|object2
collection-->|refers|object3
object1-->|owned by|collection
object2-->|owned by|collection
object3-->|owned by|collection
```

**NOTE**: The new controller is implemented as a standard [controller-runtime controller](https://book.kubebuilder.io/cronjob-tutorial/controller-implementation), which is the depart from [managed.Reconciler](https://github.com/crossplane/crossplane-runtime/blob/master/pkg/reconciler/managed/reconciler.go#L467) pattern used across provider ecosystem. Given that the new API does not create any external resources, we were able to produce simpler code. The approach was discussed and approved by @negz and @turkenh.

Fixes #209

I have:

* Added new unit tests
* deploy a test resource with `kubectl -f apply examples/collection/collection.yaml`


internal/controller/collection/reconciler.go
```diff
@@ -0,0 +1,277 @@
...
+	log.Info("Reconciling", "name", collection.Name)
+
+	// Get client for the referenced provider config.
+	clusterClient, err := r.clientForProvider(ctx, r.client, collection.Spec.ProviderConfigReference.Name)
+	if err != nil {
+		collection.Status.SetConditions(xpv1.ReconcileError(err))
+		if err := r.client.Status().Update(ctx, collection); err != nil {
+			return ctrl.Result{}, errors.Wrap(err, errStatusUpdate)
+		}
+		return ctrl.Result{}, errors.Wrap(err, errNewKubernetesClient)
```
turkenh, line 133: Looks like we are loosing the error from `r.clientForProvider` here. Overridden by the result of `r.client.Status().Update`.

internal/controller/collection/reconciler.go
```diff
@@ -0,0 +1,277 @@
...
+	objects.SetAPIVersion(collection.Spec.APIVersion)
+	objects.SetKind(collection.Spec.Kind)
+	selector, err := metav1.LabelSelectorAsSelector(&collection.Spec.Selector)
+
+	if err != nil {
+		collection.Status.SetConditions(xpv1.ReconcileError(err))
+		if err := r.client.Status().Update(ctx, collection); err != nil {
+			return ctrl.Result{}, errors.Wrap(err, errStatusUpdate)
+		}
+		return ctrl.Result{}, errors.Wrap(err, "error creating selector")
```
turkenh, line 147: Same here (and with all other error handlings), actual error lost.

internal/controller/collection/reconciler.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	// Fetch objects based on the set GVK and selector.
+	objects := &unstructured.UnstructuredList{}
+	objects.SetAPIVersion(collection.Spec.APIVersion)
+	objects.SetKind(collection.Spec.Kind)
+	selector, err := metav1.LabelSelectorAsSelector(&collection.Spec.Selector)
+
+	if err != nil {
+		collection.Status.SetConditions(xpv1.ReconcileError(err))
```
turkenh, line 143: Should use wrapped error message in the conditions (here and in all others).

collection.Status.SetConditions(xpv1.ReconcileError(errors.Wrap(err, "error creating selector")))

internal/controller/collection/reconciler.go
```diff
@@ -0,0 +1,277 @@
...
+			return ctrl.Result{}, errors.Wrap(err, errStatusUpdate)
+		}
+		return ctrl.Result{}, errors.Wrap(err, "error listing objects")
+	}
+
+	// Create observed-only Objects for all found items
+	refs := sets.New[v1alpha1.ObservedObjectReference]()
+	for _, o := range objects.Items {
+		log.Debug("creating observed object for the matched item", "gvk", o.GroupVersionKind(), "name", o.GetName())
+		name, err := r.observedObjectName(collection, &o) //#nosec G601
```
turkenh, line 162: https://github.com/crossplane/crossplane/tree/master/contributing#explain-nolint-directives

internal/controller/collection/reconciler.go
```diff
@@ -0,0 +1,277 @@
...
+		}
+	}
+	v, err := runtime.DefaultUnstructuredConverter.ToUnstructured(observedObject)
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot convert to unstructured")
+	}
+	u := &unstructured.Unstructured{Object: v}
+	u.SetGroupVersionKind(v1alpha2.ObjectGroupVersionKind)
+	u.SetName(observedObject.Name)
+	return u, nil
```
turkenh, line 276: Do we have to use unstructured client to work with `Object` type? I believe we could just work with the `Object` type without this conversion.

internal/controller/collection/reconciler.go
```diff
@@ -0,0 +1,277 @@
...
+		refs.Insert(v1alpha1.ObservedObjectReference{Name: name})
+	}
+
+	// Remove objects that either do not exist anymore or are no match
+	oldRefs := sets.New[v1alpha1.ObservedObjectReference](collection.Status.Objects...)
+	for _, or := range oldRefs.Difference(refs).UnsortedList() {
+		o := &v1alpha2.Object{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: or.Name,
+			}}
```
turkenh, line 199: Some debug logs would be helpful here, e.g. job foo no longer is a match, deleting observed resource...

internal/controller/collection/reconciler.go
```diff
@@ -0,0 +1,277 @@
...
+				return ctrl.Result{}, errors.Wrap(err, errStatusUpdate)
+			}
+			return ctrl.Result{}, errors.Wrap(err, fmt.Sprintf("cannot delete observed object %s", or.Name))
+		}
+	}
+	collection.Status.SetConditions(xpv1.ReconcileSuccess(), xpv1.Available())
+
+	// Update references only if more or less objects got matched since the last reconciliation.
+	if !refs.Equal(oldRefs) {
+		collection.Status.Objects = refs.UnsortedList()
```
turkenh, line 212: Same here, some debug logs or even throwing an event like matchingObjectsChanged would be helpful.

internal/controller/collection/reconciler.go
```diff
@@ -0,0 +1,277 @@
...
+			}
+			return ctrl.Result{}, errors.Wrap(err, "cannot create observed object")
+		}
+
+		log.Debug("created observed object", "name", po.GetName())
+		refs.Insert(v1alpha1.ObservedObjectReference{Name: name})
+	}
+
+	// Remove objects that either do not exist anymore or are no match
+	oldRefs := sets.New[v1alpha1.ObservedObjectReference](collection.Status.Objects...)
```
turkenh, line 194: Currently, we have two distinct sources (`status.objects` and objects owned by collection on etcd), and they could diverge for some reason.

I believe it would be more reliable if we lean on whatever on the API instead of using references in the status (also relying on status has its downsides). I mean, we could add labels to the created observe-only object resources (e.g. kubernetes.crossplane.io/owned-by-collection: jobs) and list them with labels to see what a collection owns and clean them if they no longer needed.

internal/controller/collection/reconciler.go
```diff
@@ -0,0 +1,277 @@
...
+			ForProvider: v1alpha2.ObjectParameters{
+				Manifest: runtime.RawExtension{
+					Raw: []byte(manifest),
+				},
+			},
+		},
+	}
+	if t := collection.Spec.Template; t != nil {
+		if len(t.Metadata.Labels) > 0 {
+			observedObject.SetLabels(t.Metadata.Labels)
```
turkenh, line 263: By default adding a label like `kubernetes.crossplane.io/owned-by-collection: jobs` would be helpful if [not necessary](https://github.com/crossplane-contrib/provider-kubernetes/pull/217/files#r1544381370). This could be used to quickly filter those object with `kubectl get` but also in the compositions.

apis/v1alpha1/types.go
```diff
@@ -114,3 +114,98 @@ type ProviderConfigUsageList struct {
...
+	Kind string `json:"kind"`
+
+	// Namespace where to look for objects.
+	// If omitted, search is performed across all namespaces.
+	// For cluster-scoped objects, omit it.
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+
+	// Selector defines the criteria for including objects into the collection
+	Selector metav1.LabelSelector `json:"selector"`
```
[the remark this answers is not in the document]
phisco: `match`?

apis/v1alpha1/types.go
```diff
@@ -114,3 +114,98 @@ type ProviderConfigUsageList struct {
 	metav1.ListMeta `json:"metadata,omitempty"`
 	Items           []ProviderConfigUsage `json:"items"`
 }
+
```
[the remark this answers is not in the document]
pedjak: sure. I thought that `apis/v1alpha1` is the current pattern.

internal/controller/collection/reconciler.go
pedjak: We need to use unstructured because we create Object using server-side apply.

internal/controller/collection/reconciler.go
pedjak: If we do not want to keep references in the status, then this becomes obsolete?

internal/clients/client.go
```diff
@@ -86,3 +106,71 @@ func restConfigFromAPIConfig(c *api.Config) (*rest.Config, error) {
 
 	return config, nil
 }
+
+// ClientForProvider returns the client for the given provider config
+func ClientForProvider(ctx context.Context, inclusterClient client.Client, providerConfigName string) (client.Client, error) { //nolint:gocyclo
```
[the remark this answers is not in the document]
pedjak: sure, initially I wanted to keep PR focus small.

internal/controller/collection/reconciler.go
pedjak: If we introduce such label, then we implicitly limit collection names to 63 chars. Let me know if this is an important trade to be made. Also, `ObservedObjectCollection` instances are gonna be created mostly from within a composition, where their name is built from composite name and a prefix.

internal/controller/collection/reconciler.go
pedjak: If we introduce such label, then we implicitly limit collection names to 63 chars. Let me know if this is an important trade to be made. Also, `ObservedObjectCollection` instances are gonna be created mostly from within a composition, where their name is built from composite name and a prefix.

We already add possibility for user to add custom labels/annotations through `.spes.template`, do we need an additional label here that is going to limit the collection name length to 63 chars?

internal/controller/collection/reconciler.go
pedjak: > I believe it would be more reliable if we lean on whatever on the API instead of using references in the status 

Having refs in the status might be helpful for some, similar to how we have refs in the composite object.

phisco: Yes, agree having refs in the status makes sense for visibility. But we should also add labels so that functions can avoid having to first get the collection and then requesting objects if they already know the collection, they should be able to request resources by label directly instead.

pedjak: >  if they already know the collection, they should be able to request resources by label directly instead.

Users can do that already by adding an arbitrary number of labels through `.spec.template.metadata.labels` on ObserverdObjectCollection.

phisco: Fair enough, then I'm fine either way.

apis/v1alpha1/types.go
```diff
@@ -114,3 +114,98 @@ type ProviderConfigUsageList struct {
...
+// ObservedObjectCollectionSpec defines the desired state of ObservedObjectCollection
+type ObservedObjectCollectionSpec struct {
+
+	// APIVersion of objects that should be matched by the selector
+	// +kubebuilder:validation:MinLength:=1
+	APIVersion string `json:"apiVersion"`
+
+	// Kind of objects that should be matched by the selector
+	// +kubebuilder:validation:MinLength:=1
+	Kind string `json:"kind"`
```
jbw976, line 154: I briefly tried to find if this was explored/concluded in the design discussion, e.g. https://github.com/crossplane-contrib/provider-kubernetes/issues/209#issuecomment-2016652897, but didn't find it after very minimal searching, so thank you for educating me @pedjak 😂 

Is the `Kind` field optional at all? i.e. does this limit the user to exactly one `Kind` of objects that this collection can find and observe? Is the expectation here that the user knows all the Kinds that an `Object` may end up creating, or that descendants of their `Object` may create (e.g., `Deployment` -> `ReplicaSet` -> `Pods`), and therefore they create an `ObservedObjectCollection` for each `Kind` they anticipate being created within that graph?

bobh66: I think the conclusion was to start with the simplest implementation which requires `Kind` and then see if we can add the option to do something more complicated later.  I think it would be really useful to be able to say "find all of the resources in the owner tree using resource X as the root" but I'm not sure how we would implement that efficiently.  The GC is able to do it and we may or may not want to/be able to replicate that functionality.  It may also be possible to write a Composition Function that has some knowledge of what a particular resource's child `Kind`s would be and use the extra-resources functionality to walk the tree and create `Object`s or `ObservedObjjectCollection`s as needed.

apis/v1alpha1/types.go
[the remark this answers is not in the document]
turkenh: Not necessarily, I see the discussion there is around the general approach on how to tackle the problem at hand rather than the final shape of API.

Given this will be our API (and it is possible that some future examples will follow the same pattern), it is worth putting more thought before committing.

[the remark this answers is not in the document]
negz: Yeah, I like the idea of nesting the fields that specify what resources to observe.

Of all the names I've seen so far, `spec.observe` sounds best to me. Maybe `spec.observeObjects`?

apis/v1alpha1/types.go
negz: I'd definitely prefer to keep it simple and require `kind`.

This is informed by a real use case from the team at Nike, which was CronJob -> Jobs and Deployment -> Pods. In both cases one input object (e.g. CronJob) will only produce one kind of object (i.e. Job) that the function is interested in.

As you mention @jbw976, if there's a reasonably small set of kinds you could always create one ObservedObjectCollection each.

internal/controller/collection/reconciler.go
```diff
@@ -0,0 +1,277 @@
...
+	// Update references only if more or less objects got matched since the last reconciliation.
+	if !refs.Equal(oldRefs) {
+		collection.Status.Objects = refs.UnsortedList()
+	}
+
+	return ctrl.Result{RequeueAfter: r.pollInterval()}, r.client.Status().Update(ctx, collection)
+}
+
+func observedObjectName(collection client.Object, matchedObject client.Object) (string, error) {
+	h := md5.New() //#nosec G401 -- used only for unique name generation
```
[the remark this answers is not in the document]
turkenh: I like the idea of leveraging MATCHED_OBJECT_UUID and avoiding extra calls to API server 👍

internal/controller/collection/reconciler.go
turkenh: Correct me if I am wrong but `r.client.Patch` with `client.Apply` should work with typed objects as well. We need unstructured only if we pass partial objects (missing some required fields) ?

internal/controller/collection/reconciler.go
turkenh: > If we introduce such label, then we implicitly limit collection names to 63 chars. 

>  Also, ObservedObjectCollection instances are gonna be created mostly from within a composition, where their name is built from composite name and a prefix.

This is a good point. However, the same limitation holds for composites, doesn't it? They put `crossplane.io/composite: <name-of-the-composite>` to all resources they compose and they could be created from within other compositions as well. 

I believe the reason why we maintain references there is that the composed resources could be of any apiVersion/kind and it is not possible to list all composed objects with a label filter. The situation is different here, we know all owned resources are `Objects` and we can list/filter them with labels. 

I am totally fine with keeping them in status for visibility.

apis/v1alpha1/types.go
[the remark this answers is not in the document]
pedjak: Thanks Nic for the suggestino - we went with `spec.observedObjects`.

internal/controller/collection/reconciler.go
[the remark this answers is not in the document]
pedjak: In order to make it backup-friendly, I have update the convention to:

```
${COLLECTION_NAME}-${MATCHED_OBJECT_KIND}-${MATCHED_OBJECT_NAMESPACE}-${MATCHED_OBJECT_NAME}
```

if the generate name is longer than 253 chars, we generate md5 hash of it and use it instead.

internal/controller/collection/reconciler.go
pedjak: Fixed, please resolve it it matches the expectations.

internal/controller/collection/reconciler.go
pedjak: I have introduced `kubernetes.crossplane.io/owned-by-collection` label that contains the name of the collection, if the name length is less than 63 chars. If greater, md5 hash of the name is used instead.

The label and its value is published under `status.membershipLabel` so that clients can know what label to use to obtain all items belonging to the given collection.

pedjak:
@turkenh I have addressed the comments, ptal.

internal/controller/collection/reconciler.go
negz: FWIW my understanding matches @pedjak's - today the reality is that if you want to use SSA you need to use `Unstructured` objects.

apis/v1alpha1/types.go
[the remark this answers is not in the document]
negz: Are you sure we want `spec.observedObjects` (i.e. observed, not observe)?

Observed is the past tense of observe. So to me `ObservedObjectCollection` makes sense as the kind, because once you declare its existence what you have is a collection of observed objects - an "observed object collection".

Past tense makes less sense to me in the context of the `spec` though, since that's where I'm telling the collection what to do - "observe these objects". So to me `spec.observeObjects` is more intuitive.

I don't feel strongly about this though, and I could just be overthinking it. 🙂

internal/controller/collection/reconciler.go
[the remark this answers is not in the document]
negz: We have a couple of existing ways to name automatically generated child resources:

* For XRs and composed resources, we use the NameGenerator, that mimics `metadata.generateName` (see [here](https://github.com/crossplane/crossplane/blob/v1.15.1/internal/names/generate.go#L68))
* For composition revisions, we use the composition name suffixed with a truncated SHA256 hash of the composition (see [here](https://github.com/crossplane/crossplane/blob/v1.15.1/internal/controller/apiextensions/composition/revision.go#L30))
* For package revisions, we use the package name suffixed with a the OCI digest (i.e. truncated SHA256 hash) of the package (see [here](https://github.com/crossplane/crossplane/blob/v1.15.1/internal/controller/pkg/manager/revisioner.go#L87))

I'd prefer not to avoid another naming scheme, unless there's strong reason to. I'd especially prefer to avoid introducing a naming scheme with two variants (e.g. sometimes one format, sometimes md5 hashes).

Can we just use `metadata.generateName`, to let the API server name these objects? This would be aesthetically identical to the composition NameGenerator, and avoid any concerns about extra calls to the API server. (Although for the record, I don't share those concerns.)

I don't think we need the NameGenerator for cases like this where an `ObservedObjectCollection` will always create one kind of resource - `Object`. It would be relatively cheap for this controller to:

* Just use `metadata.generateName`
* Proactively check for leaked resources by listing `Object` filtered by `kubernetes.crossplane.io/owned-by-collection` before creating anything new

I actually think this applies to the claim controller too - raised https://github.com/crossplane/crossplane/issues/5535 to track that.

internal/controller/collection/reconciler.go
negz: > If greater, md5 hash of the name is used instead.

FWIW there's a potential collision here:

* One collection has a long enough name that its objects refer to it as `some-md5-hash`
* Another collection is actually named `some-md5-hash`

This probably doesn't matter - it would never happen outside of a collection actively trying to cause a collision. 

Is it so bad to just limit collection names to 63 characters? I prefer predictable and human-readable labels to labels that are sometimes human readable, and other times an md5 hash. I think this would also allow us to avoid having to write a label to the status of the collection just to know what label to query for (e.g. hash or name).

internal/controller/observedobjectcollection/reconciler.go
```diff
@@ -0,0 +1,318 @@
...
+
+	"github.com/crossplane-contrib/provider-kubernetes/apis/object/v1alpha2"
+	"github.com/crossplane-contrib/provider-kubernetes/apis/observedobjectcollection/v1alpha1"
+	"github.com/crossplane-contrib/provider-kubernetes/internal/clients"
+)
+
+const (
+	errNewKubernetesClient = "cannot create new Kubernetes client"
+	errStatusUpdate        = "cannot update status"
+	fieldOwner             = client.FieldOwner("observed-object-collection-controller")
```
turkenh, line 54:
```suggestion
	fieldOwner             = client.FieldOwner("kubernetes.crossplane.io/observed-object-collection-controller")
```

nit: do we want to namespace it?

internal/controller/observedobjectcollection/reconciler.go
```diff
@@ -0,0 +1,318 @@
...
+
+	log.Info("Reconciling")
+
+	mlv, err := membershipLabelValue(c)
+	if err != nil {
+		werr := errors.Wrapf(err, "error getting value for label %v", membershipLabelKey)
+		c.Status.SetConditions(xpv1.ReconcileError(werr))
+		_ = r.client.Status().Update(ctx, c)
+		return ctrl.Result{}, werr
+	}
```
turkenh, line 134: Should we follow the same as crossplane-runtime? https://github.com/crossplane/crossplane-runtime/blob/991de68a3f6e90e2a5ee59067697048db333c0b3/pkg/reconciler/managed/reconciler.go#L903-L917

internal/controller/observedobjectcollection/reconciler.go
```diff
@@ -0,0 +1,318 @@
...
+	id, err := uuid.FromBytes(h.Sum(nil))
+	return id.String(), err
+}
+
+func membershipLabelValue(collection *v1alpha1.ObservedObjectCollection) (string, error) {
+	if len(collection.GetName()) <= 63 {
+		return collection.GetName(), nil
+	}
+	// Otherwise, compute md5 hash of it, to reduce it length.
+	h := md5.New() //#nosec G401 -- used only for unique name generation
```
turkenh, line 260: Just checked how Crossplane doing it with `crossplane.io/composite` label: https://github.com/crossplane/crossplane/blob/d35ecef8b3d9094261dbda64e72498cda1520797/internal/controller/apiextensions/composite/api.go#L401

Looks like we don't have special handling for long names rather try to add it as label, it fails and we return the error to the user so that they can adjust their naming. I remember I observed this while applying some composites (of platform-refs) and used shorter names with my compositions. 

I would agree that this is not the best experience, but still would follow upstream implementation here as well. To keep the logic here simple and not have cryptic owner labels when they are long.

internal/controller/observedobjectcollection/reconciler.go
```diff
@@ -0,0 +1,318 @@
...
+			werr := errors.Wrapf(err, "cannot delete observed object %s", name)
+			c.Status.SetConditions(xpv1.ReconcileError(werr))
+			_ = r.client.Status().Update(ctx, c)
+			return ctrl.Result{}, werr
+		}
+	}
+	c.Status.SetConditions(xpv1.ReconcileSuccess(), xpv1.Available())
+
+	// Update references only if more or less objects got matched since the last reconciliation.
+	if !refs.Equal(oldRefs) {
```
turkenh, line 229: nit: 

```suggestion
	if !refs.Equal(sets.New[v1alpha1.ObservedObjectReference](c.Status.ObjectRefs...)) {
```

and the above line, `oldRefs := sets.New[v1alpha1.ObservedObjectReference](c.Status.ObjectRefs...)` could be removed. This is the only place it is used.

internal/controller/collection/reconciler.go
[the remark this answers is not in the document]
pedjak: > Can we just use metadata.generateName, to let the API server name these objects? 

We cannot unfortunately, because a server-side apply patch must be named, i.e. `metadata.name` should be set. SSA is used because we want to propagate correctly (and remove too) annotations/labels that users can add through `spec.objectTemplate` to generated Objects.

* The proposed scheme is very simple and produces deterministic and always available names that remain stable across arbitrary number of backup/restore cycles both on crossplane and remote clusters.
* It falls back to md5 hash only if the name is longer than 253 chars (and we agree that it gonna happen rarely)

What is the advantage of using random names when we have the deterministic solution?
* There is always a chance for collision
* In each reconcile loop, we need to (1) create new objects if there are more matches, (2) remove objects not matched anymore, and (3) update all existing objects because `spec.objectTemplate` labels/annotations could change. Having the deterministic names helps here a lot to keep the logic simple.

> Proactively check for leaked resources by listing Object filtered by kubernetes.crossplane.io/owned-by-collection before creating anything new

In order to detect leakage, we need to come up with an identifier that can be found/computed on the fresh query result. If we use random Object names, we need to discover which item in the results correspond to which Object. Eventually we are going to build an index over Objects, but we could use that index key as Object name and save us one pass over them.

BTW, if we apply the random name generation starting from a claim `foo`, this is what we are going to get on collection objects:

* claim name: `foo`
* composite name: `foo-1dgft`
* observed object collection: `foo-1dgft-tr451`
* observed object collection member1: `foo-1dgft-tr451-ptrqw`
* observed object collection member1: `foo-1dgft-tr451-qwrty`

In the proposed naming schema, the last two names could look like `foo-1dgft-tr451-configmap-example1-cfg1` and `foo-1dgft-tr451-configmap-example2-cfg1`.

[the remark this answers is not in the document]
sttts: Not sure I understand all the context. What I see are two approaches:
1. two phase commit with the simple namer (including conflict recovery). This means that we store the name mapping in the collection before creating an object.
2. deterministic names. This does not need to store names. We assume the deterministic names are unique enough such that in our life-time there is never a conflict.

So it depends on whether we want to store names. I tend to prefer the deterministic approach as it is easier and has less write to the apiserver.

internal/controller/observedobjectcollection/reconciler.go
pedjak: What is the added value? The current approach:
* Indicates the real error on condition
* tries to update the status collection resource, but if it fails is fully irrelevant for reconciliation retry, controller-runtime gets the real error back
* the real error is logged [here](https://github.com/crossplane-contrib/provider-kubernetes/pull/217/files#diff-cbd20e87d322d219672285e5d13eb7d14481d19bc89343232c569801055895b8R98-R104)

Pasting the referred code here:

```golang
if err != nil {
		log.Debug("Cannot observe external resource", "error", err)
		if kerrors.IsConflict(err) {
			return reconcile.Result{Requeue: true}, nil
		}
		record.Event(managed, event.Warning(reasonCannotObserve, err))
		managed.SetConditions(xpv1.ReconcileError(errors.Wrap(err, errReconcileObserve)))
		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
	}
```

If I understand correctly, the real error is not going to be returned back to the caller in the above approach.

internal/controller/observedobjectcollection/reconciler.go
pedjak: Given that actually `status.objectRefs` is not needed because there is a common object label, and in the light of current naming debate, I am going to the refs from the status.

internal/controller/observedobjectcollection/reconciler.go
turkenh: >What is the added value? 

Not much other than following upstream. 

Sounds good to me.

internal/controller/collection/reconciler.go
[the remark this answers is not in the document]
pedjak: > two phase commit with the simple namer (including conflict recovery). This means that we store the name mapping in the collection before creating an object.

We use SSA (Patch) so we are not able to detect conflict - if object already exists, we are going to update it. Crossplane has the same issue: https://github.com/crossplane/crossplane/blob/master/internal/controller/apiextensions/composite/composition_functions.go#L473

internal/controller/collection/reconciler.go
pedjak: fixed, please resolve if it matches the expectations.

internal/controller/observedobjectcollection/reconciler.go
pedjak: I have removed `status.objectRefs` fields from CRD.

internal/controller/collection/reconciler.go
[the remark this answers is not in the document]
negz: The spirit of my feedback: The answer isn't always "what is the best thing we could do, in isolation, for this new feature".

This isn't a greenfield context. Crossplane already has two patterns to name child resources - `NameGenerator` and truncated SHA256 hashes. If we can use one of those, we should strongly consider it. Consistency matters. Consistent UX and aesthetics for end users, and consistent implementations for everyone (users and maintainers) to reason about.

If every time we face the same problem, we solve it a little bit differently (even if it's better!) we end up with a system that's more surprising and harder to reason about overall.

> Having the deterministic names helps here a lot to keep the logic simple.

That makes sense. One of the two existing patterns I called out above is deterministic - truncated SHA256 hashes.

Could we take the deterministic data you're already using (matched object GVK + namespace + name), SHA256 hash it, and truncate it to 7 characters? So for an `ObservedObjectCollection` named `jobs`, each child `Object` would have a name like `jobs-d35ecef`.

Pros:

* Deterministic names
* Uses an existing, established pattern
* Short, and collision resistant enough for at least hundreds of children of one `ObservedObjectCollection`
* Consistent format (doesn't toggle between human readable and a hash depending on input data length)

Cons:

* Not human readable

Not being human readable is a little unfortunate, but 'm not overly worried about it. The GVK, namespace and name of the wrapped object are all print columns of the `Object` CRD and will thus show up in `kubectl get objects -l kubernetes.crossplane.io/owned-by-collection=jobs`.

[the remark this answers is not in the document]
pedjak: > Could we take the deterministic data you're already using (matched object GVK + namespace + name), SHA256 hash it, and truncate it to 7 characters?

Sure, pushed the change. By looking at your earlier comments, my understanding was that you favor `NameGenerator` option.

BTW, from math point of view, using SHA256 over MD5 to compute hashes does not reduce chances for name collision, because we take just first 56 bits.

apis/observedobjectcollection/v1alpha1/types.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// ObservedObjectCollectionStatus represents the observed state of a ObservedObjectCollection
+type ObservedObjectCollectionStatus struct {
+	v12.ResourceStatus `json:",inline"`
+
+	// MembershipLabel is the label set on each member of this collection
+	// and can be used for fetching them.
+	// +optional
+	MembershipLabel map[string]string `json:"membershipLabel,omitempty"`
```
turkenh, line 116: Do we still need this field in the status? AFAIU, it is always deterministic and we don't have to store this in status.

turkenh (approved):
Thanks @pedjak 🙌

apis/observedobjectcollection/v1alpha1/types.go
pedjak: see https://github.com/crossplane-contrib/provider-kubernetes/pull/217#discussion_r1549776009

IMHO, instead of relying on an implicit convention, we can make the label discoverable by users.

lnattrass:
This is a weird question, but how is this intended to be used? :sweat_smile:

PR #134, opened 2023-08-23
Add Alpha Support for Server Side Apply

turkenh:
### Description of your changes

This PR introduces a new flag, `--enable-server-side-apply`, to enable [Server Side Apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) instead of the existing Patching Applicator combined with the last applied configuration annotation. 

Fixes #37
Fixes #114 

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

Tested/verified with `examples/object/object.yaml` and `examples/object/object-service.yaml` with/without the flag.

Verified that enabling the feature does not set last-applied configuration and properly sets field manager in the managed fields:

1. Create the `Object` example wrapping a k8s `Service`:
```bash
❯ kubectl create -f examples/object/object-service.yaml
object.kubernetes.crossplane.io/sample-service created
```
2. Check the created service with managed fields:
```bash
❯ kubectl get svc  sample-service -o yaml --show-managed-fields
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: "2023-08-23T10:29:35Z"
  labels:
    example: "true"
  managedFields:
  - apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:labels:
          f:example: {}
      f:spec:
        f:ports:
          k:{"port":80,"protocol":"TCP"}:
            .: {}
            f:port: {}
            f:protocol: {}
            f:targetPort: {}
        f:selector: {}
    manager: crossplane-provider-kubernetes
    operation: Apply
    time: "2023-08-23T10:29:35Z"
  name: sample-service
  namespace: default
  resourceVersion: "32113"
  uid: f540e8ed-4ccc-4fc2-a38f-bdb76658bc96
spec:
  clusterIP: 10.96.125.58
  clusterIPs:
  - 10.96.125.58
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - port: 80
    protocol: TCP
    targetPort: 9376
  selector:
    app.kubernetes.io/name: MyApp
  sessionAffinity: None
  type: ClusterIP
status:
  loadBalancer: {}
```
3. Make an update on the `Object`.
4. Verify the update reflected on the `Service`.

[contribution process]: https://git.io/fj2m9

Issue #242, opened 2024-05-14
Key/Value Fields Not Updating Properly


bradkwadsworth-mw:
### What problem are you facing?
When removing key/value fields, the desired manifest is not updated.
object resource:
```yaml
apiVersion: kubernetes.crossplane.io/v1alpha2
kind: Object
metadata:
  annotations:
    crossplane.io/external-create-pending: "2024-05-14T03:06:56Z"
    crossplane.io/external-create-succeeded: "2024-05-14T03:06:56Z"
    crossplane.io/external-name: sample-namespace
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"kubernetes.crossplane.io/v1alpha2","kind":"Object","metadata":{"annotations":{"uptest.upbound.io/timeout":"60"},"name":"sample-namespace"},"spec":{"forProvider":{"manifest":{"apiVersion":"v1","kind":"Namespace","metadata":{"labels":{"example":"true"}}}},"providerConfigRef":{"name":"kubernetes-provider"}}}
    uptest.upbound.io/timeout: "60"
  creationTimestamp: "2024-05-14T03:06:56Z"
  finalizers:
  - finalizer.managedresource.crossplane.io
  generation: 3
  name: sample-namespace
  resourceVersion: "30070"
  uid: 745a73ee-9447-47f7-b3c8-0105fcc2e3b4
spec:
  deletionPolicy: Delete
  forProvider:
    manifest:
      apiVersion: v1
      kind: Namespace
      metadata:
        labels:
          example: "true"
          foo: bar
  managementPolicies:
  - '*'
  providerConfigRef:
    name: kubernetes-provider
  readiness:
    policy: SuccessfulCreate
  watch: false
```
created resource:
```yaml
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: '{"apiVersion":"v1","kind":"Namespace","metadata":{"labels":{"example":"true","foo":"bar"}}}'
  creationTimestamp: "2024-05-14T03:06:56Z"
  labels:
    example: "true"
    foo: bar
    kubernetes.io/metadata.name: sample-namespace
  name: sample-namespace
  resourceVersion: "31736"
  uid: 7174f6ef-af55-4ae2-bc81-0f6fef75384b
spec:
  finalizers:
  - kubernetes
```

object resource (removed label):
```yaml
[...]

Issue #213, opened 2024-03-08
Release scope for v1.0.0 


jeanduplessis:
The following issues are planned for inclusion in the v1.0.0 release of this provider:

```[tasklist]
### Tasks
- [ ] https://github.com/crossplane-contrib/provider-kubernetes/issues/37
- [ ] https://github.com/crossplane-contrib/provider-kubernetes/issues/114
- [ ] https://github.com/crossplane-contrib/provider-kubernetes/issues/159
- [ ] https://github.com/crossplane-contrib/provider-kubernetes/issues/212
- [ ] https://github.com/crossplane-contrib/provider-kubernetes/issues/209
```

jeanduplessis:
@bobh66 unfortunately not. It was a case of prioritizing the XPv2 work above other features in this provider. I'll discuss with maintainers what can be done to push these topics forward.

PR #246, opened 2024-05-16
Add Alpha Support for Server Side Apply (2nd attempt)

### LINKED ISSUE #37: failure to detect drift in created kubernetes resources
### What happened?
Kubernetes Deployment created from provider-kubernetes object is not monitored for drift.

### How can we reproduce it?

Create the following object:
```yaml
kind: Object
metadata:
  name: whoami
  namespace: default
spec:
  forProvider:
    manifest:
      apiVersion: apps/v1
      kind: Deployment
      metadata:
        namespace: default
      spec:
        metadata:
          name: whoami
          namespace: default
        replicas: 2
        selector:
          matchLabels:
            app: whoami
        template:
          metadata:
            labels:
              app: whoami
          spec:
            containers:
            - name: whoami
              image: docker.io/containous/whoami:v1.5.0
  providerConfigRef:
    name: kubernetes-provider
```

Edit the created deployment with `kubectl edit deployments.apps whoami` updating `spec.replicas: 3`

Expected: deployment drift detected and updated back to `spec.replicas: 2`.  
Actual: Deployment continues to have 3 pods.

Editing the object `forProvider.manifest.spec.replicas: 4`:
Expected: object drift detected and updated back to `forProvider.manifest.spec.replicas: 2` and deployment and pods updated.
Actual: object replicas, deployment replicas and pod count remains at 4

Note that if the object is created using composition, it behaves as above on deployment edit, BUT then if the object is updated `forProvider.manifest.spec.replicas: 4` then the drift of object is somehow detected, the replicas reset to 2 and the deployment and pod count similarly fixed up back to 2.

```yaml
apiVersion: apiextensions.crossplane.io/v1
kind: CompositeResourceDefinition
metadata:
  name: xminapps.example.com
spec:
  group: example.com
  names:
    kind: XMinApp
    plural: xminapps
  claimNames:
    kind: MinAppClaim
    plural: minappclaims
  versions:
  - name: v1alpha1
    served: true
    referenceable: true
    schema:
      openAPIV3Schema:
        type: object
        properties:
          spec:
            type: object
            properties:
              id:
                type: string
              parameters:
                type: object
                properties:
                  imageName:
                    type: string
                  replicas:
                    type: integer
                    default: 1
                required:
                - imageName
            required:
            - id
            - parameters
---
apiVersion: apiextensions.crossplane.io/v1
kind: Composition
metadata:
  name: minapp
  labels:
    crossplane.io/xrd: xminapp.example.com
    provider: kubernetes
spec:
  compositeTypeRef:
    apiVersion: example.com/v1alpha1
    kind: XMinApp
  resources:
  - name: deployment
    base:
      kind: Object
      spec:
        forProvider:
          manifest:
            apiVersion: apps/v1
            kind: Deployment
            metadata:
              namespace: default
            spec:
              replicas: 1
              template:
                spec:
                  containers:
                  - name: frontend
        providerConfigRef:
          name: kubernetes-provider
    patches:
    - fromFieldPath: spec.id
      toFieldPath: metadata.name
    - fromFieldPath: spec.id
      toFieldPath: spec.forProvider.manifest.metadata.name
    - fromFieldPath: spec.id
      toFieldPath: spec.forProvider.manifest.metadata.labels.app
    - fromFieldPath: spec.id
      toFieldPath: spec.forProvider.manifest.spec.selector.matchLabels.app
    - fromFieldPath: spec.id
      toFieldPath: spec.forProvider.manifest.spec.template.metadata.labels.app
    - fromFieldPath: spec.id
      toFieldPath: spec.forProvider.manifest.spec.template.spec.containers[0].name
    - fromFieldPath: spec.parameters.imageName
      toFieldPath: spec.forProvider.manifest.spec.template.spec.containers[0].image
    - fromFieldPath: spec.parameters.replicas
      toFieldPath: spec.forProvider.manifest.spec.replicas
---
apiVersion: example.com/v1alpha1
kind: MinAppClaim
metadata:
  name: my-whoami
spec:
  id: whoami
  parameters:
    imageName: docker.io/containous/whoami:v1.5.0
    replicas: 2
```

Crossplane version: v1.7.0 (helm install)
kubernetes version v1.22.5 (windows docker desktop kubernetes environment)
Win 10 21H2 (19044.1645)
accessing through through WSL2 Ubuntu 20.04.4 LTS (Focal Fossa)
using client kubectl v1.21.5

### LINKED ISSUE #57: Patch existing resource in merge patch way
Scenario:
I have EKS cluster and I need to update aws-node daemonset environment variable.
I try to do it with following Object definition, this is part of composition:
```
  - name: aws-node-daemonset
    base:
      kind: Object
      spec:
        managementPolicy: ObserveCreateUpdate
        providerConfigRef:
          name: kubernetes-provider
        forProvider:
          manifest:
            apiVersion: apps/v1
            kind: DaemonSet
            metadata:
              name: aws-node
              namespace: kube-system
            spec:
              template:
                spec:
                  containers:
                  - env:
                    - name: ENABLE_PREFIX_DELEGATION
                      value: "true"
                    name: aws-node
```
Result is that I get error `update failed: cannot apply object: cannot patch object: DaemonSet.apps "aws-node" is invalid: spec.template.spec.containers[0].image: Required value`

So it looks like I have to add complete daemonset definition to make it work but that's not optimal, I just want to modify existing a little bit.
Is there anyway to do some kind of "merge patch" to existing kubernetes object? I think that way this could work

### LINKED ISSUE #114: Provider breaks kubectl 3-way diff apply
### What happened?

I'm having crossplane patch an existing manifest. For some reason crossplane is replacing the last-applied-configuration annotation with it's own one. This breaks the kubernetes 3-way diff, and leads to all kinds of surprises.
Breaking 3-way diff for kubectl apply is a major violation of k8s concepts IMHO.
The provider should strictly stick to the server-side apply managed fields mechanism instead.

Please at least provide an option to opt-out from the kubernetes provider writing that annotation.

### How can we reproduce it?

```yaml
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: test
  namespace: default
  labels:
    test: "yes"
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: test-annotation-patches
  namespace: default
data:
  iam_role_arn: "arn:aws:iam::111122223333:role/test-annotation-role"
```

I'm apply-ing the above manifests.

Then I introduce this Object to apply an annotation from the configmap to the ServiceAccount:

```yaml
---
kind: Object
metadata:
  name: test
spec:
  managementPolicy: ObserveCreateUpdate
  forProvider:
    manifest:
      apiVersion: v1
      kind: ServiceAccount
      metadata:
        namespace: default
  references:
  - patchesFrom:
      apiVersion: v1
      kind: ConfigMap
      name: test-annotation-patches
      namespace: default
      fieldPath: data[iam_role_arn]
    toFieldPath: metadata.annotations[eks.amazonaws.com/role-arn]
  providerConfigRef:
    name: kubernetes-provider
```

When I then delete the `test` label from the service-account I'd expect the 3-way kubernetes diff to deal with the deletion of the annotation, but this is how it looks like after crossplane has applied the patch:

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  annotations:
    eks.amazonaws.com/role-arn: arn:aws:iam::111122223333:role/test-annotation-role
    kubectl.kubernetes.io/last-applied-configuration: '{"apiVersion":"v1","kind":"ServiceAccount","metadata":{"annotations":{"eks.amazonaws.com/role-arn":"arn:aws:iam::111122223333:role/test-annotation-role"},"namespace":"default"}}'
  creationTimestamp: "2023-05-08T10:57:58Z"
  labels:
    test: "yes"
  name: test
  namespace: default
  resourceVersion: "647740"
  uid: 76795938-a2aa-4b4a-96d0-f6a56620bd13
```

When looking at the `Object` I'm seeing the following:

```yaml
kind: Object
metadata:
  annotations:
    crossplane.io/external-name: test
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"kubernetes.crossplane.io/v1alpha1","kind":"Object","metadata":{"annotations":{},"name":"test"},"spec":{"forProvider":{"manifest":{"apiVersion":"v1","kind":"ServiceAccount","metadata":{"namespace":"default"}}},"managementPolicy":"ObserveCreateUpdate","providerConfigRef":{"name":"kubernetes-provider"},"references":[{"patchesFrom":{"apiVersion":"v1","fieldPath":"data[iam_role_arn]","kind":"ConfigMap","name":"test-annotation-patches","namespace":"default"},"toFieldPath":"metadata.annotations[eks.amazonaws.com/role-arn]"}]}}
  creationTimestamp: "2023-05-08T11:03:30Z"
  finalizers:
  - finalizer.managedresource.crossplane.io
  generation: 2
  name: test
  resourceVersion: "647741"
  uid: 9fe7e34a-6848-4766-97fe-3c5ec8554db9
spec:
  deletionPolicy: Delete
  forProvider:
    manifest:
      apiVersion: v1
      kind: ServiceAccount
      metadata:
        annotations:
          eks.amazonaws.com/role-arn: arn:aws:iam::111122223333:role/test-annotation-role
        namespace: default
  managementPolicy: ObserveCreateUpdate
  providerConfigRef:
    name: kubernetes-provider
  references:
  - patchesFrom:
      apiVersion: v1
      fieldPath: data[iam_role_arn]
      kind: ConfigMap
      name: test-annotation-patches
      namespace: default
    toFieldPath: metadata.annotations[eks.amazonaws.com/role-arn]
status:
  atProvider:
    manifest:
      apiVersion: v1
      kind: ServiceAccount
      metadata:
        annotations:
          eks.amazonaws.com/role-arn: arn:aws:iam::111122223333:role/test-annotation-role
          kubectl.kubernetes.io/last-applied-configuration: '{"apiVersion":"v1","kind":"ServiceAccount","metadata":{"annotations":{"eks.amazonaws.com/role-arn":"arn:aws:iam::111122223333:role/test-annotation-role"},"namespace":"default"}}'
        creationTimestamp: "2023-05-08T10:57:58Z"
        labels:
          test: "yes"
        managedFields:
        - apiVersion: v1
          fieldsType: FieldsV1
          fieldsV1:
            f:metadata:
              f:annotations: {}
              f:labels:
                .: {}
                f:test: {}
          manager: kubectl-client-side-apply
          operation: Update
          time: "2023-05-08T11:16:43Z"
        - apiVersion: v1
          fieldsType: FieldsV1
          fieldsV1:
            f:metadata:
              f:annotations:
                f:eks.amazonaws.com/role-arn: {}
                f:kubectl.kubernetes.io/last-applied-configuration: {}
          manager: crossplane-kubernetes-provider
          operation: Update
          time: "2023-05-08T11:17:30Z"
        name: test
        namespace: default
        resourceVersion: "647740"
        uid: 76795938-a2aa-4b4a-96d0-f6a56620bd13
  conditions:
  - lastTransitionTime: "2023-05-08T11:03:30Z"
    reason: ReconcileSuccess
    status: "True"
    type: Synced
  - lastTransitionTime: "2023-05-08T11:03:30Z"
    reason: Available
    status: "True"
    type: Ready
```

I have no idea why the provider wants to track the `last-applied-configuration` annotation, but IMHO it should stick to the managed fields mechanism and server-side apply.

Crossplane version: 1.12.1
Provider version: 0.7.0
Kubernetes distribution: Kind
Kubernetes Sever version 1.26.3
Kubernetes Client version: 1.25.6

### LINKED ISSUE #115: New management policy: ObserveUpdate
There are some cases when you just want to patch a resource created by some other tool.

You deploy with `kubectl` directly, or your CD tool deploys the resource, and Crossplane composition patches it the first time it sees the resource, otherwise it should not create the resource, also it should not delete the resource.

In k8s it's common that it is not a single tool managing a resource, but multiple ones manage certain aspects of it.

The provider could work in a way where it only patches existing resources if they have been created, but otherwise just don't create the resource.

With the current management policies this is not possible really, An `ObserveUpdate` policy is missing.

### LINKED ISSUE #145: k8s configMap doesn't change
### What happened?
EKS ConfigMap doesn't not change 

### How can we reproduce it?
As part of provisioning of a new EKS cluster we want to modify the `aws-auth` configMap in order to map SSO IAM roles there. 
The Composite looks like the following
```yaml
    - base:
        kind: Object
        spec:
          deletionPolicy: Orphan
          providerConfigRef: {}
          forProvider:
            manifest:
              apiVersion: v1
              kind: ConfigMap
              metadata:
                name: aws-auth
                namespace: kube-system
              data:
                mapAccounts: ""
                mapRoles: ""
                mapUsers: ""
      name: eks-aws-auth-config
      patches:
        - fromFieldPath: metadata.name
          toFieldPath: spec.providerConfigRef.name
          policy:
            fromFieldPath: Required    
    
        - fromFieldPath: spec.parameters.awsAuth.mapAccounts
          toFieldPath: spec.forProvider.manifest.data.mapAccounts
        - fromFieldPath: spec.parameters.awsAuth.mapUsers
          toFieldPath: spec.forProvider.manifest.data.mapUsers

        - type: CombineFromComposite
          toFieldPath: spec.forProvider.manifest.data.mapRoles
          combine:
            variables:
            - fromFieldPath: status.ngRoleArn
            - fromFieldPath: spec.parameters.awsAuth.mapRoles
            strategy: string
            string:
              fmt: |
                - rolearn: %s
                  username: system:node:{{EC2PrivateDNSName}}
                  groups:
                  - system:bootstrappers
                  - system:nodes
                %s
```
Everything works well, we can connect the cluster after provisioning but we noticed that if we try to modify the configMap manually, Crossplane Kubernetes provider doesn't override this change

So, let's assume that we have the following configMap in the new cluster
```yaml
k get configmap -n kube-system aws-auth -o yaml
apiVersion: v1
data:
  mapAccounts: ""
  mapRoles: |+
    - rolearn: arn:aws:iam::***:role/hellofresh-eks-cbtc5
      username: system:node:{{EC2PrivateDNSName}}
      groups:
      - system:bootstrappers
      - system:nodes
    - groups:
      - system:masters
      rolearn: arn:aws:iam::***:role/***
      username: ***

  mapUsers: ""
kind: ConfigMap
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: '{"apiVersion":"v1","data":{"mapAccounts":"","mapRoles":"-
      rolearn: arn:aws:iam::***:role/hellofresh-eks-cbtc5\n  username: system:node:{{EC2PrivateDNSName}}\n  groups:\n  -
      system:bootstrappers\n  - system:nodes\n- groups:\n  - system:masters\n  rolearn:
      arn:aws:iam::***:role/***\n  username:
      sso-eks-test-admin\n\n","mapUsers":""},"kind":"ConfigMap","metadata":{"name":"aws-auth","namespace":"kube-system"}}'
  creationTimestamp: "2023-10-09T09:11:35Z"
  name: aws-auth
  namespace: kube-system
  resourceVersion: "1029"
  uid: a3b0ed75-ef77-461d-bb87-2da000ff0201
```
and as soon as we modify it(set up wrong group for the second role)
```bash
k edit configmaps -n kube-system aws-auth      
configmap/aws-auth edited

k get configmap -n kube-system aws-auth -o yaml
apiVersion: v1
data:
  mapAccounts: ""
  mapRoles: |+
    - rolearn: arn:aws:iam::***:role/hellofresh-eks-cbtc5
      username: system:node:{{EC2PrivateDNSName}}
      groups:
      - system:bootstrappers
      - system:nodes
    - groups:
      - system
      rolearn: arn:aws:iam::***:role/***
      username: ***

  mapUsers: ""
kind: ConfigMap
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: '{"apiVersion":"v1","data":{"mapAccounts":"","mapRoles":"-
      rolearn: arn:aws:iam::***:role/hellofresh-eks-cbtc5\n  username: system:node:{{EC2PrivateDNSName}}\n  groups:\n  -
      system:bootstrappers\n  - system:nodes\n- groups:\n  - system:masters\n  rolearn:
      arn:aws:iam::***:role/***\n  username:
      sso-eks-test-admin\n\n","mapUsers":""},"kind":"ConfigMap","metadata":{"name":"aws-auth","namespace":"kube-system"}}'
  creationTimestamp: "2023-10-09T09:11:35Z"
  name: aws-auth
  namespace: kube-system
  resourceVersion: "47148"
  uid: a3b0ed75-ef77-461d-bb87-2da000ff0201
```

the controller sees this change
```
k get object hellofresh-eks-bcmc4  -o yaml
kind: Object
metadata:
  annotations:
    crossplane.io/composition-resource-name: eks-aws-auth-config
    crossplane.io/external-name: hellofresh-eks-bcmc4
  creationTimestamp: "2023-10-09T08:58:43Z"
  finalizers:
  - finalizer.managedresource.crossplane.io
  generateName: hellofresh-eks-
  generation: 3
  labels:
    crossplane.io/claim-name: ""
    crossplane.io/claim-namespace: ""
    crossplane.io/composite: hellofresh-eks
  name: hellofresh-eks-bcmc4
  ownerReferences:
  - apiVersion: aws.hellofresh.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: eks
    name: hellofresh-eks
    uid: 9d89b96f-dd1b-4887-85a9-79718a30bfb7
  resourceVersion: "12976254"
  uid: 445c8ee7-6b5b-455f-aee0-8d772978bbc4
spec:
  deletionPolicy: Orphan
  forProvider:
    manifest:
      apiVersion: v1
      data:
        mapAccounts: ""
        mapRoles: |+
          - rolearn: arn:aws:iam::***:role/hellofresh-eks-cbtc5
            username: system:node:{{EC2PrivateDNSName}}
            groups:
            - system:bootstrappers
            - system:nodes
          - groups:
            - system:masters
            rolearn: arn:aws:iam::***:role/***
            username: ***

        mapUsers: ""
      kind: ConfigMap
      metadata:
        name: aws-auth
        namespace: kube-system
  managementPolicy: Default
  providerConfigRef:
    name: hellofresh-eks
  readiness:
    policy: SuccessfulCreate
status:
  atProvider:
    manifest:
      apiVersion: v1
      data:
        mapAccounts: ""
        mapRoles: |+
          - rolearn: arn:aws:iam::***:role/hellofresh-eks-cbtc5
            username: system:node:{{EC2PrivateDNSName}}
            groups:
            - system:bootstrappers
            - system:nodes
          - groups:
            - system
            rolearn: arn:aws:iam::***:role/***
            username: ***

        mapUsers: ""
      kind: ConfigMap
      metadata:
        annotations:
          kubectl.kubernetes.io/last-applied-configuration: '{"apiVersion":"v1","data":{"mapAccounts":"","mapRoles":"-
            rolearn: arn:aws:iam::***:role/hellofresh-eks-cbtc5\n  username:
            system:node:{{EC2PrivateDNSName}}\n  groups:\n  - system:bootstrappers\n  -
            system:nodes\n- groups:\n  - system:masters\n  rolearn: arn:aws:iam::***:role/***\n  username:
            ***\n\n","mapUsers":""},"kind":"ConfigMap","metadata":{"name":"aws-auth","namespace":"kube-system"}}'
        creationTimestamp: "2023-10-09T09:11:35Z"
        managedFields:
        - apiVersion: v1
          fieldsType: FieldsV1
          fieldsV1:
            f:data: {}
          manager: vpcLambda
          operation: Update
          time: "2023-10-09T09:11:35Z"
        - apiVersion: v1
          fieldsType: FieldsV1
          fieldsV1:
            f:data:
              f:mapAccounts: {}
              f:mapUsers: {}
            f:metadata:
              f:annotations:
                .: {}
                f:kubectl.kubernetes.io/last-applied-configuration: {}
          manager: crossplane-kubernetes-provider
          operation: Update
          time: "2023-10-09T09:11:45Z"
        - apiVersion: v1
          fieldsType: FieldsV1
          fieldsV1:
            f:data:
              f:mapRoles: {}
          manager: kubectl-edit
          operation: Update
          time: "2023-10-09T13:03:50Z"
        name: aws-auth
        namespace: kube-system
        resourceVersion: "47148"
        uid: a3b0ed75-ef77-461d-bb87-2da000ff0201
  conditions:
  - lastTransitionTime: "2023-10-09T09:11:45Z"
    reason: ReconcileSuccess
    status: "True"
    type: Synced
  - lastTransitionTime: "2023-10-09T09:11:45Z"
    reason: Available
    status: "True"
    type: Ready
```

but doesn't revert it. I assume it's because the last applied configuration doesn't represent this change.

So, either we don't really understand the logic behind this new concept or something goes wrong

Crossplane Version: v1.13.2
AWS provider versions: v0.41.0
Kubernetes provider version: v0.9.0
Kubernetes Version: 1.25
Kubernetes Distribution: EKS

turkenh:

This PR adds alpha support for Server Side Apply, renovating the [previous attempt](https://github.com/crossplane-contrib/provider-kubernetes/pull/134).

Fixes #37
Fixes #57
Fixes #114 
Fixes #115
Fixes #145

I have:

#### Automated

```
UPTEST_EXAMPLE_LIST="examples/object/object-ssa-owner.yaml" make e2e
```

#### Manuel

1. Enable server side apply.
2. Apply `examples/object/object-ssa-owner.yaml`
3. Apply `examples/object/object-ssa-labeler.yaml`
4. Ensure the created service doesn't have `last-applied-annotation`.
5. Ensure the created service has labels from both `Objects`.
6. Ensure the created service has managedFields set properly.

```yaml
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: "2024-05-16T18:08:16Z"
  labels:
    another-key: another-value
    some-key: some-value
  managedFields:
  - apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:labels:
          f:some-key: {}
      f:spec:
        f:ports:
          k:{"port":80,"protocol":"TCP"}:
            .: {}
            f:port: {}
            f:protocol: {}
            f:targetPort: {}
        f:selector: {}
    manager: provider-kubernetes/sample-service-owner
    operation: Apply
    time: "2024-05-16T18:08:15Z"
  - apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:labels:
          f:another-key: {}
    manager: provider-kubernetes/sample-service-labeler
    operation: Apply
    time: "2024-05-16T18:08:26Z"
  name: sample-service
  namespace: default
  resourceVersion: "1142"
  uid: a7abcf8a-ae1d-427d-8df9-a9e53a4b2ce9
spec:
  clusterIP: 10.96.47.67
  clusterIPs:
  - 10.96.47.67
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - port: 80
    protocol: TCP
    targetPort: 9376
  selector:
    app.kubernetes.io/name: MyApp
  sessionAffinity: None
  type: ClusterIP
status:
  loadBalancer: {}
```


internal/controller/object/object.go
```diff
@@ -282,11 +302,39 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
...
+		// In an ideal world, we should be able to compare the extracted observation
+		// with what we will apply as desired state, however, due to the poor
+		// handling of defaults in the server-side apply, we cannot do that,
+		// since we always see a diff between the extracted state and the desired
+		// state in that case. This dry run call returns what we will see on
+		// the object including the defaults at the cost of one extra call to
+		// the apiserver, so that we can compare it with the extracted state to
+		// decide whether the object is up-to-date or not.
+		desiredObj := desired.DeepCopy()
+		if err := c.client.Patch(ctx, desiredObj, client.Apply, client.ForceOwnership, client.FieldOwner(ssaFieldOwner(cr.Name)), client.DryRunAll); err != nil {
```
turkenh, line 328: This a workaround for https://github.com/kubernetes/kubernetes/issues/115563

internal/controller/object/object.go
```diff
@@ -282,11 +302,39 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
...
 	}
 
-	var last *unstructured.Unstructured
-	if last, err = getLastApplied(cr, observed); err != nil {
+	// observation contains the extracted state of the observed object that
+	// should be compared with the desired state of the object manifest to
+	// decide whether the object is up-to-date or not.
+	// If serverSideApply is enabled, we will extract the state from the
+	// observed object, otherwise we will extract the state from the last
+	// applied annotation.
```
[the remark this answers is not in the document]
turkenh: > Is the extracted state only the fields of the object that are owned by the SSA field manager/owner?

Yes, correct. Reworded accordingly.

internal/controller/object/object.go
```diff
@@ -326,6 +381,13 @@ func (c *external) Update(ctx context.Context, mg resource.Managed) (managed.Ext
 		return managed.ExternalUpdate{}, err
 	}
 
+	if c.ssaExtractor != nil {
+		if err = c.client.Patch(ctx, obj, client.Apply, client.ForceOwnership, client.FieldOwner(ssaFieldOwner(cr.Name))); err != nil {
```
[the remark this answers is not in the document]
turkenh: I am not willing to do that TBH, at least with this PR. 
For now, I would like to leave it as a manual migration for those upgrading from the client side and consider implementing it later, depending on the feedback.

internal/controller/object/object.go
```diff
@@ -302,12 +350,19 @@ func (c *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 		return managed.ExternalCreation{}, err
 	}
 
+	if c.ssaExtractor != nil {
```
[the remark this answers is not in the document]
turkenh: I don't think this will make things better since there will be lots of duplication in that case. 

While it is not perfect because of the if branches in its current shape, I found it more readable since it only impacts where we calculate the desired state and do the applies.

internal/controller/object/object.go
```diff
@@ -282,11 +302,17 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
...
 	}
 
-	var last *unstructured.Unstructured
-	if last, err = getLastApplied(cr, observed); err != nil {
+	// observation contains the extracted state of the observed object that
+	// should be compared with the desired state of the object manifest to
+	// decide whether the object is up-to-date or not.
+	// If serverSideApply is enabled, we will extract the state from the
+	// observed object, otherwise we will extract the state from the last
+	// applied annotation.
```
sttts, line 310: who owns defaults? How safe is a comparison here? Don't know the logic enough of the provider. What is this value used for?

internal/controller/object/indexes.go
```diff
@@ -123,8 +123,10 @@ func enqueueObjectsForReferences(ca cache.Cache, log logging.Logger) func(ctx co
 		}
 		// queue those Objects for reconciliation
 		for _, o := range objects.Items {
-			log.Info("Enqueueing Object because referenced resource changed", "name", o.GetName(), "referencedGVK", rGVK.String(), "referencedName", ev.Object.GetName(), "providerConfig", pc)
-			q.Add(reconcile.Request{NamespacedName: types.NamespacedName{Name: o.GetName()}})
+			if o.Spec.Watch {
```
sttts, line 126: why is this necessary?

internal/controller/object/object.go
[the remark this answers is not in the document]
negz: I have two concerns here:

* Adding branches to almost every method of the ExternalClient that toggle its behavior.
* The condition of the branch - the presence of an `ssaExtractor`.

I feel two less complex implementation is better than one more complex implementation that has two different modes of operation. Each less complex implementation is easier to reason about, test, etc in isolation. Looking at this `ExternalClient`, it seems like `Observe` is the only method with much complexity right now - the rest I wouldn't be concerned about duplicating.

Using `ssaExtractor` to determine whether to SSA or works, but you're using a side effect of SSA being enabled in one place to determine whether SSA is enabled in another place. The nilness of the extractor is being used not only to gate whether extraction happens, but whether SSA happens at all.

If this was c/c I'd push back harder on this direction, but give I'm not a maintainer on this repo I'm okay to disagree and go with this more pragmatic approach.

internal/controller/object/object.go
```diff
@@ -312,6 +312,28 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
...
+		desiredObj := desired.DeepCopy()
+		if err := c.client.Patch(ctx, desiredObj, client.Apply, client.ForceOwnership, client.FieldOwner(ssaFieldOwner(cr.Name)), client.DryRunAll); err != nil {
+			return managed.ExternalObservation{}, errors.Wrap(CleanErr(err), "cannot dry run SSA")
+		}
+		desired, err = c.ssaExtractor.Extract(desiredObj, ssaFieldOwner(cr.Name))
+		if err != nil {
+			return managed.ExternalObservation{}, errors.Wrap(err, "cannot extract SSA")
+		}
+	}
+
```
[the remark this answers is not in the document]
turkenh: > re-apply our desired state on every event

That wouldn't work since, in some cases, re-applying the same manifest simply [bumps the resourceVersion](https://github.com/kubernetes/kubernetes/issues/124605) and causing another update event. So, if we implement it in that way, we would end up in an update loop, i.e., apply -> update event -> apply -> update event ...

There are various issues that I believe all boil down to the same root cause:

- https://github.com/kubernetes/kubernetes/issues/124605
- https://github.com/kubernetes/kubernetes/issues/121404
- https://github.com/kubernetes/kubernetes/issues/124050
- https://github.com/kubernetes/kubernetes/issues/115563

> If we have a apiserver roundtrip anyway for dry-run

So, if the above bug didn't exist, we wouldn't need that dry-run. This is already a workaround for that.

internal/controller/object/indexes.go
turkenh: We only want to enqueue the Object if `spec.watch` is set to true. Not every Object wants to watch the referenced resources. If you're questioning why we have such a field in the first place, I have a comment for that [here](https://github.com/crossplane-contrib/provider-kubernetes/pull/235#discussion_r1589157066).

internal/controller/object/object.go
turkenh: which defaults do you mean? 

Basically, in Observe, we are trying to decide whether the resource is up to date or not. Prior to SSA, we were using "kubectl.kubernetes.io/last-applied-configuration" to see which manifest was applied and if it is different than the current one we were deciding that we need an update. Obviously, this has a good amount of downsides (see the issues closed with this PR) and now we are trying to do the same with SSA in a more proper way. We are using unstructured extractor to extract the observed state and will use it to make the comparison. However, currently SSA is not working as advertised due to issues [here](https://github.com/crossplane-contrib/provider-kubernetes/pull/246#discussion_r1606334812) and I had to use a workaround with dry run to extract the observed state.

internal/controller/object/object.go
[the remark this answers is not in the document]
negz: > So, if we implement it in that way, we would end up in an update loop, i.e., apply -> update event -> apply -> update event

Given that composition functions uses SSA to apply composed resources, should we expect the same problem in c/c if realtime compositions are used together with functions?

internal/controller/object/indexes.go
sttts: Not questioning the watch field. But why we make the `q.Add` conditional on it. I understand that we don't start watches on their referenced resources, but here we are not doing that, but queuing them.

internal/controller/object/object.go
sttts: Same question as discussed in https://github.com/crossplane-contrib/provider-kubernetes/pull/246#discussion_r1605251719.

internal/controller/object/object.go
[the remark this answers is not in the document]
turkenh: Probably. 

I reproduced with kubectl using [the deployment in this issue](https://github.com/kubernetes/kubernetes/issues/124050), and know that an empty map triggers it. So, it doesn't happen for all resources but for some. I am almost sure there will be some MRs resulting in similar behavior under certain conditions. [This](https://github.com/kubernetes/kubernetes/issues/124605) is reproducing with a CRD / CR for example.

internal/controller/object/object.go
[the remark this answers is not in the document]
turkenh: I still don't like having multiple external client implementations due to duplication, but I see your point.

Just pushed a [commit](https://github.com/crossplane-contrib/provider-kubernetes/pull/246/commits/c6d5a72468ae4afc0a3568e2349465f958888ec8) hopefully resolving both concerns, basically by having two different implementations of a new interface, one for old patch-based and another for SSA. This way, I could avoid branches without duplication.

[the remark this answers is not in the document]
turkenh: I still need to fix unit tests ...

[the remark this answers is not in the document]
negz: Thanks, the new approach LGTM.

internal/controller/object/object.go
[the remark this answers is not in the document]
negz: That's quite unfortunate. 🙁 I've added some notes to the SSA and realtime composition lifecycle tracking issues just in case we see this in the wild.

internal/controller/object/indexes.go
turkenh: Ah, I see. 

Watches started based on GVK not including their namespaced name, so, another object with `spec.watch: true` might have started a watch on that given GVK. For example, think two Objects; one referencing `secret: foo` with `watch: true`, another one referencing `secret: bar` with `watch: false`. Even though the latter does not start watch on Secrets, it was started with the former. So, we need to filter out here, while adding to the queue otherwise the latter object will be getting watch events even not desired.

bobh66:
@turkenh @negz @sttts Where does this PR stand?  We are seeing problems similar to #242 and #114 so if there is any way to get this into a release for alpha testing that would be great.  Thanks

turkenh:
> @turkenh @negz @sttts Where does this PR stand? We are seeing problems similar to #242 and #114 so if there is any way to get this into a release for alpha testing that would be great. Thanks

@bobh66 just pushed a commit wrapping the work up with best effort. I believe the PR is ready to go.

@sttts opened a follow up [issue](https://github.com/crossplane-contrib/provider-kubernetes/issues/269) to avoid the dry-run calls we discussed above. We will make sure to resolve it before promoting the feature to beta (and/or shipping [v1.0.0](https://github.com/crossplane-contrib/provider-kubernetes/issues/213))

cmd/provider/main.go
```diff
@@ -66,6 +66,7 @@ func main() {
 
 		enableManagementPolicies = app.Flag("enable-management-policies", "Enable support for Management Policies.").Default("true").Envar("ENABLE_MANAGEMENT_POLICIES").Bool()
 		enableWatches            = app.Flag("enable-watches", "Enable support for watching resources.").Default("false").Envar("ENABLE_WATCHES").Bool()
+		enableServerSideApply    = app.Flag("enable-server-side-apply", "Enable server side apply to sync object manifests to k8s API.").Default("false").Envar("ENABLE_SERVER_SIDE_APPLY").Bool()
```
sttts, line 69: Is this meant as a feature gate? Doesn't look like from the name. I don't like to force ppl to understand so low-level concepts, i.e. this kind of flag should not exist medium-term.

internal/controller/object/object.go
```diff
@@ -247,50 +279,73 @@ func (c *connector) Connect(ctx context.Context, mg resource.Managed) (managed.E
...
+		syncer: &PatchingResourceSyncer{
+			client: resource.ClientApplicator{
+				Client:     k,
+				Applicator: resource.NewAPIPatchingApplicator(k),
+			},
+		},
+	}
+
+	if c.ssaEnabled {
+		dc, err := discovery.NewDiscoveryClientForConfig(rc)
```
sttts, line 291: This looks wrong. It is called on every reconcile, isn't it? Instantiating a new discovery client is heavy.

turkenh: The problem is, this is not something static, we need to initialize discovery client for the kubeconfig configured in the ProviderConfig referenced in the Object. It can change anytime, so, we need to initialize during reconcile.

cmd/provider/main.go
turkenh: What is wrong with name, what could be an alternative? 

I agree that people shouldn't need to understand details on how we sync the desired state, but don't feel comfortable to replace existing mechanism in place, so wanted to put behind a feature flag. I think similar to https://github.com/crossplane/crossplane/blob/2a427fc89da4a2e86f75eaecfa8328f53e19f33b/cmd/crossplane/core/core.go#L115

internal/controller/object/object.go
sttts: We have been digging a bit deeper. The extractor below downloads the OpenAPI v2 schema via `dc`. This is a big (multi-megabyte) proto file. It will grow even more (a lot) if you have many CRDs from providers installed. And the reconciler will do that on every reconcile, i.e. tens-of-megabyte download and unmarshalling. This is likely a blocker for the approach 🔴.

The solution might be caching of the discovery client by provider config.

bobh66:
@turkenh is there a way to move this forward?

turkenh:
> @turkenh is there a way to move this forward?

@bobh66 we are on it. 
@erhancagirici is [working](https://github.com/turkenh/provider-kubernetes/pulls) on resolving the two points blocking this PR.

internal/controller/object/object.go
erhancagirici: https://github.com/turkenh/provider-kubernetes/pull/3 is merged into this feature branch, that implements a custom CachingUnstructuredExtractor, which utilizes OpenAPI v3 discovery and caching per GV.

cmd/provider/main.go
sttts: If this is a feature gate that goes away after alpha/beta, then this is fine. Wasn't sure whether it is meant to stay forever as a knob.

pkg/kube/client/ssa/caching_unstructured_extractor.go
```diff
@@ -0,0 +1,237 @@
...
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
+	"k8s.io/kube-openapi/pkg/handler3"
+	"k8s.io/kube-openapi/pkg/schemamutation"
+	"k8s.io/kube-openapi/pkg/spec3"
+	"k8s.io/kube-openapi/pkg/validation/spec"
+)
+
+// cachingUnstructuredExtractor is a caching implementation of v1.UnstructuredExtractor
+// using OpenAPI V3 discovery information.
```
sttts, line 28: be clear what it caches.

pkg/kube/client/ssa/gvk_parser_cache.go
```diff
@@ -0,0 +1,76 @@
...
+	return c
+}
+
+// LoadOrNewCacheForProviderConfig returns the *GVKParserCache for the given provider config,
+// initializing an empty cache for the first use.
+// the implementation is concurrency-safe.
+func (cm *GVKParserCacheManager) LoadOrNewCacheForProviderConfig(pc *v1alpha1.ProviderConfig) (*GVKParserCache, error) {
+	cm.mu.Lock()
+	defer cm.mu.Unlock()
+	sc, ok := cm.cacheStore[pc.GetUID()]
```
sttts, line 45: why is the uid enough? Couldn't the content change?

pkg/kube/client/ssa/openapi_groupversion.go
```diff
@@ -0,0 +1,91 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/kube-openapi/pkg/handler3"
+)
+
+// OpenAPIGroupVersion is the context-aware variant of openapi.GroupVersion
+// It also stores the ETag for the underlying discovery GV path
+type OpenAPIGroupVersion interface {
+	// Schema is the context-accepting variant of the upstream client-go implementation
+	Schema(ctx context.Context, contentType string) ([]byte, error)
+	ETag() string
```
sttts, line 37: I don't understand why a schema loader and an arbitrary static etag are combined in one interface / struct.

pkg/kube/client/ssa/state_cache.go
```diff
@@ -0,0 +1,107 @@
...
+	xpresource "github.com/crossplane/crossplane-runtime/pkg/resource"
+
+	objectv1alpha2 "github.com/crossplane-contrib/provider-kubernetes/apis/object/v1alpha2"
+)
+
+// StateCacheManager lets you manage StateCache entries for XP managed
+// resources
+type StateCacheManager interface {
+	LoadOrNewForManaged(mg xpresource.Managed) StateCache
+	Remove(mg xpresource.Managed)
```
sttts, line 20: Is there any connection of this to the openapi/extractor caching? Can we put them into different packages?

pkg/kube/client/ssa/gvk_parser_cache.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	return c
+}
+
+// LoadOrNewCacheForProviderConfig returns the *GVKParserCache for the given provider config,
+// initializing an empty cache for the first use.
+// the implementation is concurrency-safe.
+func (cm *GVKParserCacheManager) LoadOrNewCacheForProviderConfig(pc *v1alpha1.ProviderConfig) (*GVKParserCache, error) {
+	cm.mu.Lock()
+	defer cm.mu.Unlock()
```
sttts, line 44: it's a RWMutex. We should first try the happy/fast path with RLock and only on fail try the slow path with full lock.

pkg/kube/client/ssa/gvk_parser_cache.go
```diff
@@ -0,0 +1,76 @@
...
+type GVKParserCache struct {
+	mu sync.RWMutex
+	// Parsers per GroupVersion
+	store map[schema.GroupVersion]*GVKParserCacheEntry
+}
+
+// GVKParserCacheEntry wraps the *GvkParser with an ETag for
+// freshness check against discovery data
+type GVKParserCacheEntry struct {
+	parser *GvkParser
```
sttts, line 74: GVKParser

pkg/kube/client/ssa/gvk_parser_cache.go
```diff
@@ -0,0 +1,76 @@
...
+	defer cm.mu.Unlock()
+	delete(cm.cacheStore, pc.GetUID())
+}
+
+// GVKParserCache holds the cached parser instances and the ETags
+// of the associated provider config.
+// Parsers are generated and cached per GroupVersion
+type GVKParserCache struct {
+	mu sync.RWMutex
+	// Parsers per GroupVersion
```
sttts, line 67: this comment does not add anything.

pkg/kube/client/ssa/gvk_parser_cache.go
```diff
@@ -0,0 +1,76 @@
...
+	store map[schema.GroupVersion]*GVKParserCacheEntry
+}
+
+// GVKParserCacheEntry wraps the *GvkParser with an ETag for
+// freshness check against discovery data
+type GVKParserCacheEntry struct {
```
sttts, line 73: any reason this is public? Do we leak it outside of the package?

pkg/kube/client/ssa/gvk_parser_cache.go
erhancagirici: cache manager stores the collection of caches per provider config.
The actual cache is the one that is returned, this is done at Connect time and given to the reconciliation scope.
Then the content changes are handled there.

pkg/kube/client/ssa/gvk_parser.go
```diff
@@ -0,0 +1,129 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+This file is forked from upstream kubernetes/apimachinery
+https://github.com/kubernetes/apimachinery/blob/2465dc5239ab8827a637148a78b380c278b4a5f4/pkg/util/managedfields/gvkparser.go
```
sttts, line 19: why is it forked? What differs?

pkg/kube/client/ssa/gvk_parser_cache.go
```diff
@@ -0,0 +1,76 @@
...
+	"sync"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/crossplane-contrib/provider-kubernetes/apis/v1alpha1"
+)
+
+// GVKParserCacheManager maintains GVK parser cache stores for each provider config.
+type GVKParserCacheManager struct {
```
sttts, line 17: // It is thread-safe.

pkg/kube/client/ssa/caching_unstructured_extractor.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	oapiPathsToGV := map[string]OpenAPIGroupVersion{}
+	for path, oapiGV := range discoMap.Paths {
+		parse, err := url.Parse(oapiGV.ServerRelativeURL)
+		if err != nil {
+			return nil, err
+		}
+		useClientPrefix := strings.HasPrefix(oapiGV.ServerRelativeURL, "/openapi/v3")
+		etag := parse.Query().Get("hash")
```
sttts, line 82: is hash considered an interface in Kube ? Can it be renamed to etag tomorrow?

pkg/kube/client/ssa/gvk_parser_cache.go
erhancagirici: Left it like that intentionally, to match the [upstream implementation](https://github.com/kubernetes/apimachinery/blob/2465dc5239ab8827a637148a78b380c278b4a5f4/pkg/util/managedfields/gvkparser.go#L35). I can change it.

pkg/kube/client/ssa/caching_unstructured_extractor.go
```diff
@@ -0,0 +1,237 @@
...
+	if !ok {
+		return nil, fmt.Errorf("cannot find GroupVersion %q in discovery", gvPath)
+	}
+
+	// check the cache after invalidating stale data
+	parserTuple, ok := e.cache.store[gv]
+	// cache hit
+	if ok && parserTuple.etag == oapiGV.ETag() && oapiGV.ETag() != "" {
+		return parserTuple.parser, nil
+	}
```
sttts, line 117:
```suggestion
	if ok && parserTuple.etag == oapiGV.ETag() && oapiGV.ETag() != "" {
		// cache hit
		return parserTuple.parser, nil
	}
```

pkg/kube/client/ssa/caching_unstructured_extractor.go
```diff
@@ -0,0 +1,237 @@
...
+	// generate new parser on cache miss or etag mismatch
+	// defensively cover the case where discovery does not return any ETag
+	// for GV, which normally should not happen
+	freshParser, err := newParserFromOpenAPIGroupVersion(ctx, oapiGV)
```
sttts, line 121: is this expensive? I.e. is it doing any client requests? If yes, we are blocking the cache here.

In other words: why is the mutex a RWMutex if we don't use that?

pkg/kube/client/ssa/caching_unstructured_extractor.go
```diff
@@ -0,0 +1,237 @@
...
+	if err != nil {
+		return nil, err
+	}
+	// cache parser only if non-empty etag
```
sttts, line 125: doesn't add anything

pkg/kube/client/ssa/caching_unstructured_extractor.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+// ExtractStatus is the same as ExtractUnstructured except
+// that it extracts the status subresource applied configuration.
+// Experimental!
+func (e *cachingUnstructuredExtractor) ExtractStatus(object *unstructured.Unstructured, fieldManager string) (*unstructured.Unstructured, error) {
+	return e.extractUnstructured(object, fieldManager, "status")
+}
+
+func discoveryPaths(ctx context.Context, rc rest.Interface) (map[string]OpenAPIGroupVersion, error) {
```
sttts, line 59: this seems to be misnamed. I don't see it returning a path. I see it returning schema loaders by paths.

Why is it indexed by paths in the first place? Why not directly by gv? It is only used with gvs.

pkg/kube/client/ssa/caching_unstructured_extractor.go
```diff
@@ -0,0 +1,237 @@
...
+	data, err := rc.Get().
+		AbsPath("/openapi/v3").
+		Do(ctx).
+		Raw()
+
```
sttts, line 64:
```suggestion
	data, err := rc.Get().AbsPath("/openapi/v3").Do(ctx).Raw()
```

No empty lines before error handling.

pkg/kube/client/ssa/state_cache.go
erhancagirici: Yes, we can. There is no direct relation, but they are related in the context of provider-kubernetes's SSA-enabled client

pkg/kube/client/ssa/caching_unstructured_extractor.go
```diff
@@ -0,0 +1,237 @@
...
+		// note: this should not mutate any ref, only validate
+		RefCallback: validateRefSelfContainedFn(&refErrors, oapi.Components.Schemas),
+	}
+	specs := map[string]*spec.Schema{}
+	for k, v := range oapi.Components.Schemas {
+		walker.WalkSchema(v)
+		specs[k] = v
+	}
+	if len(refErrors) > 0 {
+		return nil, errors.Errorf("cannot validate references in OpenAPI schemas: %s", strings.Join(refErrors, ",\n"))
```
sttts, line 175: why not use kube's `NewAggregate` with `[]error`. Why this manual string acrobatic?

internal/controller/object/object.go
sttts: The original blocker problem is solved with the caching.

pkg/kube/client/ssa/gvk_parser_cache.go
sttts:
```
At connect time you call `LoadOrNewCacheForProviderConfig`, but it will return the same object on every reconcile. If the ProviderConfig is changes, the UID will still be the same and with that the cache is potentially invalid. What do I miss?

File: pkg/kube/client/ssa/state_cache.go (hunk 13 above)
Comment (sttts) [reply pkg/kube/client/ssa/state_cache.go:20]: ssa/cache/state
ssa/cache/extractor

Would make it easier to understand.

File: pkg/kube/client/ssa/gvk_parser_cache.go (hunk 14 above)
Comment (sttts) [reply pkg/kube/client/ssa/gvk_parser_cache.go:44]: then at least make it normal mutex

File: pkg/kube/client/ssa/gvk_parser_cache.go (hunk 11 above)
Comment (erhancagirici) [reply pkg/kube/client/ssa/gvk_parser_cache.go:45]: > At connect time you call LoadOrNewCacheForProviderConfig, but it will return the same object on every reconcile.
yes, we get the same GvkParserCache here for the PC. `LoadOrNewCacheForProviderConfig` just gives you the actual cache to maintain. We can think of `GvkParserCacheManager` like a mere storage organizer of caches, it just points you to the right drawer reserved for PC. Then we pass the GvkParserCache we got to the extractor at Connect time.

The cache itself is maintained after that point by the extractor [here](https://github.com/crossplane-contrib/provider-kubernetes/pull/246/files#diff-a3b611789875d8619ff3bd58cacbf0cea61cd467e8fcdb294936358addb0cd8eR98)

at each extract request:
A v3 discovery is initiated that returns a collection of all served gv paths to Etagged API paths, then according to the output :
- stale entries are invalidated (non-matching Etags, APIs not being served anymore) 
- then we retrieve the schema (if it is not on the cache after invalidations) for the relevant GV of object that is being extracted, and cache it with its ETag.

So, in the case of the ProviderConfig change, the returned discovery information will change, as the discovery client itself is rebuilt at each reconcile. Then, the invalidations will take place, so we operate on up-to-date info.

File: pkg/kube/client/ssa/gvk_parser.go (hunk 18 above)
Comment (erhancagirici) [reply pkg/kube/client/ssa/gvk_parser.go:19]: was forked due to build a new parser without the proto.Models (as building a proto.Models from v3 doc had an upstream issue).
but realized that the only difference we introduce is about NewGVKParser, no need to redefine the struct itself. Will change 👍

File: pkg/kube/client/ssa/gvk_parser_cache.go (hunk 11 above)
Comment (sttts) [reply pkg/kube/client/ssa/gvk_parser_cache.go:45]: Ack. Seems sounds. This is crucial information that needs to be in some comments

File: pkg/kube/client/ssa/caching_unstructured_extractor.go (hunk 24 above)
Comment (erhancagirici) [reply pkg/kube/client/ssa/caching_unstructured_extractor.go:59]: terminology and most parts of the implementation was borrowed from upstream [client-go](https://github.com/kubernetes/client-go/blob/fb6ca040d7b7e2293898e45dc0bfd9181f553bcc/openapi/client.go#L43)

I think naming of `openapi.GroupVersion` is a bit confusing too. The underlying implementor is a  [handler3.OpenAPIDiscoveryGroupVersion](https://github.com/kubernetes/kube-openapi/blob/9e1beecbcb384a484fb5cbd41024f316d51fdfac/pkg/handler3/handler.go#L57) that is a not-necessarily-GV server path wrapped along with a rest client.
Below is the response of``kubectl get --raw '/openapi/v3'`
The returned paths are not necessarily GroupVersion, it includes Groups, and other non-GV paths.
Probably this is why it is called `Paths()` at upstream.

Can try to parse the paths and filter the GV ones, then store `schema.GroupVersion` as keys. Though calculating the path keys from GVs seemed simpler.

```json
{
  "paths": {
    ".well-known/openid-configuration": {
      "serverRelativeURL": "/openapi/v3/.well-known/openid-configuration?hash=73BCD53F499D27D225D70A0C79D84C6D96EF410A3E0B49E471D7CDB055230C486083EF11FC1C29E6AEACAF25B2D2B7E49297FE432F15A419E5991D503203C7F7"
    },
    "api": {
      "serverRelativeURL": "/openapi/v3/api?hash=180F6A8D2E4479013F92230D4093DE33088F2E1DA0C220298FBA4ED8129D327D3C24DFE5BBEA4945457A8A117FC948E89F3630AA06EDA06A868875A3DBE567FF"
    },
    "api/v1": {
      "serverRelativeURL": "/openapi/v3/api/v1?hash=2A0B7D1DD0AF45569217219A34EF7E93C9AA2026A039B7C199C8677779832FB2D69D1CB6DA68A0EBE54F0EFDDB2DB2ABFC47CE8DF8213B52ED5BF31C9AC42BC4"
    },
    "apis": {
      "serverRelativeURL": "/openapi/v3/apis?hash=9331F4E2B1319F29C18C93095ED9BF8790E36D9D56F8B1956CF3487402BF394D50DACB1FE91BBBAD05E15AD7B45AA524B76AADB67F243F33E454456842D5B053"
    },
    "apis/admissionregistration.k8s.io": {
      "serverRelativeURL": "/openapi/v3/apis/admissionregistration.k8s.io?hash=C45DE744067E28F6002B35EB8C3B9DDB574FAC161809AB3C864C3D5BE6593C5D02D4649C8570A94731C5ABAE848AE53324EFEAE97691F85ED88BBC5B190AC04C"
    },
    "apis/admissionregistration.k8s.io/v1": {
      "serverRelativeURL": "/openapi/v3/apis/admissionregistration.k8s.io/v1?hash=9FDF107E3C61B36E079CE1A262B2EB39437937BDB37CDE8052EE41A5FCAB091BC40ACE0AFF31EF209BEA0E8E473669641A93C0D6D6A5F12E8797978950D25B4B"
    },
    "...." :{
      "...": "REMOVED FOR BREVITY"
    },
    "logs": {
      "serverRelativeURL": "/openapi/v3/logs?hash=AC0F20C36395B3CF2711AE80C53F4304EA2F3951102EBF89A6B5E9F6AEAB60E1FF879C4384177B7AA61A890512A4806AB967C143579087A17D8C1CC29006153E"
    },
    "openid/v1/jwks": {
      "serverRelativeURL": "/openapi/v3/openid/v1/jwks?hash=FEB706252E8F3767A0134706D62378FCC2C589BFD1E9F11DB562F0B783AD4B4065ED5938AEC3D49302B8B0168D86538A13A475F4FCB5B5DCB96D419A4A30F59C"
    },
    "version": {
      "serverRelativeURL": "/openapi/v3/version?hash=329C6D6D665A9D8059D57E79584EBD07DD8EAC487FE776AE43627595767FD6C43ECEF82DFC3A30884961A18803AAAF5AB2C7F9E898E7A9410E31ECD1D9AC9EEA"
    }
  }
}
```

pkg/kube/client/ssa/caching_unstructured_extractor.go
erhancagirici: Yes, it is doing a client request to `/openapi/v3/group/version` to retrieve a single GV schema.

Again, it was implemented with a `RLock()` then `Lock()` when cache miss. Then, intentionally switched to single Lock per request at review, to keep all cache related operations in single critical section accepting the performance penalty, assuming cache misses are relatively rare after first call.
I'll switch to regular mutex to keep it consistent.

pkg/kube/client/ssa/caching_unstructured_extractor.go
erhancagirici: not an interface in exact terms, but the upstream [implementation](https://github.com/kubernetes/kube-openapi/blame/9e1beecbcb384a484fb5cbd41024f316d51fdfac/pkg/handler3/handler.go#L128) and [comment](https://github.com/kubernetes/kube-openapi/blame/9e1beecbcb384a484fb5cbd41024f316d51fdfac/pkg/handler3/handler.go#L58) seems to imply that way. Anyways, this implementation in general is hopefully will be an interim solution.

sttts: Upstream is built in a way that the client does not need to know the path format. It might be that we get broken here. Chance is not super high, but possible. Alternative: just store the full path with query params.

pkg/kube/client/ssa/caching_unstructured_extractor.go
sttts: Not good. If there are external requests, this can block. Seconds, if you are unlucky and there is no timeout: minutes.

pkg/kube/client/ssa/openapi_groupversion.go
erhancagirici: Ideally, we wanted to extract the ETag from the schema loader's HTTP response headers and use it for caching. 
So, something like
```
Schema(ctx context.Context, contentType string) (data []byte, etag string, err error)
```
However,  client-go `rest.Result` does not expose the response headers. 

Therefore, we "optimistically" associate the ETag that we got from the discovery with the schema loader, since we query the schema right after. In the case of, the schema content (the real ETag) changes in the short time between setting etag and loading schema, from functional correctness perspective we are not affected by storing with a "outdated etag" in the cache. We'll be using the latest correct schema,  but we'll just cause cache miss in the next reconciliation, since we always do discovery first and invalidate stales and re-query.
`ETag()` was interfaced per suggestion from previous review cycle. It makes sense to me to have it on another map, so that we associate them in our logic, not in the struct/interface.

Other alternative considered was to always query for Schema with `If-None-Match` header set to the extracted hash and rely on the API server returning 304, that hints us to use cached instance. This approach still requires extracting hash from discovery ofc.

pkg/kube/client/ssa/caching_unstructured_extractor.go
erhancagirici: agreed on the concern, but for the alternative, AFAIU if the server is changed in a way that it does not return any kind of hash-like query param, we risk unintentionally caching with a non-unique key, therefore operating on stale schema.

In the current state of the PR, we'll skip caching and we will only lose the caching ability, in case we cannot find `hash` in query params, but we continue operating with correct schema.

WDTY about an follow-up issue for an integration test implementation that checks whether discovery returns `hash`?

pkg/kube/client/ssa/caching_unstructured_extractor.go
erhancagirici: done https://github.com/crossplane-contrib/provider-kubernetes/pull/246/commits/81c421cac3d398c3688895af32e0c6fcde1a288e

pkg/kube/client/ssa/caching_unstructured_extractor.go
erhancagirici: done, https://github.com/crossplane-contrib/provider-kubernetes/pull/246/commits/e8b85762a2218c877855bfba9c28cd5d4bb401a6 uses RLock() to check first, then uses `singleflight` to group schema fetch requests, and ensures only one in-flight request per GV. Concurrent requests are therefore deduplicated, also this happens outside of lock.

pkg/kube/client/ssa/state_cache.go
erhancagirici: done

pkg/kube/client/ssa/openapi_groupversion.go
erhancagirici: done https://github.com/crossplane-contrib/provider-kubernetes/pull/246/commits/9f522106676864f4c7527dffe1cf8533d8eb1434, Etag is removed from the interface and we return them as a separate map

pkg/kube/client/ssa/caching_unstructured_extractor.go
sttts: Isn't that just a 5 liner? Just do it.

pkg/kube/client/ssa/cache/extractor/openapi_groupversion.go
```diff
@@ -0,0 +1,84 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+This file is forked from upstream kubernetes/client-go
+https://github.com/kubernetes/client-go/blob/0b9a7d2f21befcfd98bf2e62ae68ea49d682500d/openapi/groupversion.go
```
sttts, line 19: why?

examples/object/testhooks/validate-ssa.sh
```diff
@@ -0,0 +1,25 @@
...
+
+if ! ${KUBECTL} get service sample-service -o jsonpath='{.metadata.annotations}' | grep -v "last-applied-configuration"; then # This annotation should not be present when SSA is enabled
+  echo "SSA validation failed! Annotation 'last-applied-configuration' should not exist when SSA is enabled!"
+  #exit 1
+fi
+if ! (${KUBECTL} get service sample-service -o jsonpath='{.metadata.labels.some-key}' | grep -q "some-value" && ${KUBECTL} get service sample-service -o jsonpath='{.metadata.labels.another-key}' | grep -q "another-value"); then
+  echo "SSA validation failed! Labels 'some-key' and 'another-key' from both Objects should exist with values 'some-value' and 'another-value' respectively!"
+  #exit 1
+fi
+echo "Successfully validated the SSA feature!"
```
sttts, line 19: is this the only e2e test we have? 😱

turkenh: Currently, we rely on the uptest framework for e2e tests just like other providers and what we can do with uptest is limited, but still better than nothing. Also due to some current [limitations](https://github.com/crossplane/uptest/issues/17) that is about to be solved, we couldn't really integrate this test into the CI.

I would suggest:
- Merging this PR by manually triggering those tests.
- Fast follow to enable them in CI with the corresponding uptest fix, so that we can make sure they run for every PR.
- Creating a ticket to discuss whether we need a more sophisticated testing framework here.

Issue #192, opened 2024-02-02
Extend the readinessChecks Policy options to allow for a list of Condition types and values to match


bobh66:
### What problem are you facing?
Many Kubernetes Custom Resources have `status.conditions` that should be ignored for readiness checks so the `AllTrue` policy will not work.

### How could Crossplane help solve your problem?
It would be useful to be able to specify a list of condition types and associated values to match on that determine resource readiness

phisco:
We kind of had this planned as an extension of the FromObject readiness policy, see [here](https://github.com/crossplane-contrib/provider-kubernetes/pull/119#discussion_r1203461505). But it feels you'd like for it to be more powerful and be able to match multiple conditions but not all of them, right?

bobh66:
Yes - I'm dealing with resources that might have a dozen or more status conditions but only a few of them are relevant to resource readiness, so I need to be able to filter on the condition types that are considered.

PR #247, opened 2024-05-18
feat: adds derive from cel query readiness policy

zach-source:

Adds new readiness policy called "DeriveFromCelQuery". This policy allows the use of a cel query to check if an object is ready. This is handy to check conditions on objects with non-standard methods of determining if it ready such as the argocd applictation object.

Probably closes https://github.com/crossplane-contrib/provider-kubernetes/issues/192

I have:

I've written several test caes that cover the usage of the new feature.


bobh66:
@zach-source this looks great - do you have an example of how to reference a status condition with a specific `type` in a CEL query?  I have been looking at CEL examples and I haven't found an equivalent syntax to the JSONPath `$.status.conditions[?(@.type=="Completed")].status` syntax.

bobh66:
> @zach-source this looks great - do you have an example of how to reference a status condition with a specific `type` in a CEL query? I have been looking at CEL examples and I haven't found an equivalent syntax to the JSONPath `$.status.conditions[?(@.type=="Completed")].status` syntax.

I think I figured it out:

status.conditions.exists(w, w.type=='Foo' && w.status == 'True')

zach-source:
Ah yea, exactly. And you can combine them if you want multiple. I've pushed up changes to make it pass all the reviewable checks. It's been a minute since I worked on a provider and I always have a hard time getting that submodule so sync correctly.

bobh66:
@zach-source Please run `go mod tidy` and `make generate` and commit the results.

zach-source:
Should be good to go now @bobh66

zach-source:
Had to break up the function as gocyclo got a little angry at me.

apis/object/v1alpha2/types.go
```diff
@@ -131,6 +134,16 @@ type Readiness struct {
 	// +kubebuilder:validation:Enum=SuccessfulCreate;DeriveFromObject;AllTrue
```
bobh66, line 134: I think this Enum needs to be updated with the new value

zach-source: Oh heck - good catch. I'll update it and push here shortly

bobh66:
@turkenh can you take a look at this?  It would be great to get this into the next release if possible - thanks!

internal/controller/object/object.go
```diff
@@ -479,6 +472,103 @@ func getReferenceInfo(ref v1alpha2.Reference) (string, string, string, string) {
...
+)
+
+// checkDeriveFromCelQuery will look at the celQuery field and run it as a program, using the observed object as input to
+// evaluate if the object is ready or not
+func (c *external) checkDeriveFromCelQuery(obj *v1alpha2.Object, observed *unstructured.Unstructured) (ready bool, message string) {
+	if obj.Spec.Readiness.CelQuery == "" {
+		c.logger.Debug("cel query is empty")
+		message = "cel query cannot be empty"
+		return ready, message
+	}
```
phisco, line 521: this could be done at the CRD level, to avoid erroring out at runtime 🤔

internal/controller/object/object.go
```diff
@@ -479,6 +472,103 @@ func getReferenceInfo(ref v1alpha2.Reference) (string, string, string, string) {
...
+
+	env, _ := cel.NewEnv(
```
phisco, line 523: any particular reason not to handle the err here?

internal/controller/object/object.go
```diff
@@ -479,6 +472,103 @@ func getReferenceInfo(ref v1alpha2.Reference) (string, string, string, string) {
...
+		}
+	}
+	return
+}
+
+const (
+	msgCelQueryFailedToCompile       = "failed to compile query; %s"
+	msgCelQueryReturnTypeNotBool     = "celQuery does not return a bool type; %s"
+	msgCelQueryFailedToCreateProgram = "failed to create program from the cel query; %s"
+	msgCelQueryFailedToEvalProgram   = "failed to eval the program; %s"
```
phisco, line 511: nit: all these are mostly used to wrap errors, so just make them `errFmt*` consts above, make the function below return an error instead, use `errors.Wrap` to add the original error if needed, and unwrap it when setting it as message. WDYT?

zach-source: Sure.

internal/controller/object/object.go
zach-source: Old age?

internal/controller/object/object.go
zach-source: Good idea. Should be a validation. Will add.

zach-source:
> Looking good! A few minor comments

Thanks for the review - I'll do a bit of work on it this weekend and push it up. Probably tomorrow.

internal/controller/object/object.go
zach-source: I've kept this block. I'm worried the cel validation at the CRD level doesn't kick in to all supported versions of kubernetes yet. Anyways always good to guard just in case.

zach-source:
@phisco comments have been addressed.

internal/controller/object/object.go
```diff
@@ -483,6 +488,105 @@ func getReferenceInfo(ref v1alpha2.Reference) (string, string, string, string) {
...
+	}
+	allTrue = len(conditioned.Conditions) > 0
+	for _, condition := range conditioned.Conditions {
+		if condition.Status != v1.ConditionTrue {
+			allTrue = false
+			break
+		}
+	}
+	return
+}
```
phisco, line 521: nit: we try to avoid naked returns, as they make code more obscure.
```suggestion
func (c *external) checkAllConditions(observed *unstructured.Unstructured) bool {
	conditioned := xpv1.ConditionedStatus{}
	err := fieldpath.Pave(observed.Object).GetValueInto("status", &conditioned)
	if err != nil {
		c.logger.Debug("Got error while getting conditions from observed object, setting it as Unavailable", "error", err, "observed", observed)
		return false
	}
	allTrue = len(conditioned.Conditions) > 0
	for _, condition := range conditioned.Conditions {
		if condition.Status != v1.ConditionTrue {
			return false
		}
	}
	return true
}
```

internal/controller/object/object.go
```diff
@@ -420,39 +431,33 @@ func (c *external) setObserved(obj *v1alpha2.Object, observed *unstructured.Unst
 func (c *external) updateConditionFromObserved(obj *v1alpha2.Object, observed *unstructured.Unstructured) error {
 	switch obj.Spec.Readiness.Policy {
```
phisco, line 432: nit: here we could declare ready and err before hand, just set them in the switch and move the code setting the conditions to after that.
```golang
func (c *external) updateConditionFromObserved(obj *v1alpha2.Object, observed *unstructured.Unstructured) error {
	var ready bool
	var err error
	switch obj.Spec.Readiness.Policy {
	case v1alpha2.ReadinessPolicyDeriveFromObject:
		ready = c.checkDeriveFromObject(observed)
	case v1alpha2.ReadinessPolicyAllTrue:
		ready = c.checkAllConditions(observed)
	case v1alpha2.ReadinessPolicyDeriveFromCelQuery:
		ready, err = c.checkDeriveFromCelQuery(obj, observed)
	case v1alpha2.ReadinessPolicySuccessfulCreate, "":
		// do nothing, will be handled by c.handleLastApplied method
		// "" should never happen, but just in case we will treat it as SuccessfulCreate for backward compatibility
	default:
		// should never happen
		return errors.Errorf("unknown readiness policy %q", obj.Spec.Readiness.Policy)
	}

	if err != nil {
		obj.SetConditions(xpv1.Unavailable().WithMessage(err.Error()))
		return nil
	}

	if !ready {
		obj.SetConditions(xpv1.Unavailable())
		return nil
	}

	obj.SetConditions(xpv1.Available())
	return nil
}
```

internal/controller/object/object.go
```diff
@@ -483,6 +488,105 @@ func getReferenceInfo(ref v1alpha2.Reference) (string, string, string, string) {
...
+		c.logger.Debug("Got error while getting conditions from observed object, setting it as Unavailable", "error", err, "observed", observed)
+		return
+	}
+	if status := conditioned.GetCondition(xpv1.TypeReady).Status; status != v1.ConditionTrue {
+		c.logger.Debug("Observed object is not ready, setting it as Unavailable", "status", status, "observed", observed)
+		return
+	}
+	ready = true
+	return
+}
```
phisco, line 504: nit: we try to avoid naked returns, as they make code more obscure.
```suggestion
func (c *external) checkDeriveFromObject(observed *unstructured.Unstructured) bool {
	conditioned := xpv1.ConditionedStatus{}
	if err := fieldpath.Pave(observed.Object).GetValueInto("status", &conditioned); err != nil {
		c.logger.Debug("Got error while getting conditions from observed object, setting it as Unavailable", "error", err, "observed", observed)
		return false
	}
	if status := conditioned.GetCondition(xpv1.TypeReady).Status; status != v1.ConditionTrue {
		c.logger.Debug("Observed object is not ready, setting it as Unavailable", "status", status, "observed", observed)
		return false
	}
	return true
}
```

phisco (approved):
Looking good! just a few small nits, we can address them in a follow up pr if needed

internal/controller/object/object.go
zach-source: Looks cleaner. I'm a fan.

zach-source:
@phisco Added your recommended changes/nits.

PR #251, opened 2024-05-22
Add support for auth with Upbound identity and refactor kube client builder as an importable package

turkenh:

This PR adding support for authenticating with Upbound identity using session/robot tokens.

With the number of identity providers increasing, I took the liberty of refactoring the relevant code piece so that we can consume the same package from the provider-helm (and other similar providers) instead of duplicating the code. After this PR, I'll open a PR to consume the package `github.com/crossplane-contrib/provider-kubernetes/pkg/kube/` from provider-helm. It may not be the perfect solution but I believe it is a step forward and we can easily move it to a shared place (e.g. crossplane-runtime) if we decide to do so.

I have:

Configure & Create: 

```bash
export UPBOUND_ACCOUNT=<YOUR_ACCOUNT>
export UPBOUND_TOKEN=<YOUR_API_TOKEN> # Robot tokens will be supported soon
export CONTROLPLANE_CONFIG=/tmp/controlplane-kubeconfig
rm -f $CONTROLPLANE_CONFIG

up alpha web-login -a $UPBOUND_ACCOUNT
KUBECONFIG=$CONTROLPLANE_CONFIG up ctx $UPBOUND_ACCOUNT/upbound-gcp-us-west-1/default/hasan-test

kubectl -n crossplane-system create secret generic cluster-config --from-file=kubeconfig=$CONTROLPLANE_CONFIG
kubectl -n crossplane-system create secret generic upbound-credentials --from-literal=token=$UPBOUND_TOKEN

kubectl apply -f examples/provider/provider-config-with-secret-upbound-identity.yaml
kubectl apply -f examples/object/object.yaml
```

Observe:

```bash
❯ kubectl get object
NAME               KIND        PROVIDERCONFIG        SYNCED   READY   AGE
sample-namespace   Namespace   kubernetes-provider   True     True    4s

❯ KUBECONFIG=$CONTROLPLANE_CONFIG kubectl get ns
NAME                STATUS   AGE
crossplane-system   Active   32h
default             Active   32h
kube-node-lease     Active   32h
kube-public         Active   32h
kube-system         Active   32h
sample-namespace    Active   4m20s
upbound-system      Active   32h
```


internal/clients/upbound/upbound.go
```diff
@@ -0,0 +1,86 @@
...
+		client: auth.NewClient(&up.Config{
+			Client: up.NewClient(func(client *up.HTTPClient) {
+				// TODO: This should probably be configurable.
+				client.BaseURL.Host = "auth.upbound.io"
+			}),
+		}),
+		org:          org,
+		refreshToken: string(token),
+	}
+	ts := oauth2.ReuseTokenSource(t, us)
```
turkenh, line 59: This is executed in each `Connect,` and we initialize a reuse token source with an initial token of `nil`, which refreshes/exchanges tokens before they actually expire (in each reconcile instead if once in an hour).  It looks like the issue already exists with other identity methods as well, e.g. Google and Azure, so we should fix them for all.

turkenh: This is resolved by introducing a `ReuseTokenSource Store` initialized during Setup.

pkg/kube/client/upbound/upbound.go
```diff
@@ -0,0 +1,92 @@
...
+// tokens fetched using the supplied Upbound session/robot token.
+func WrapRESTConfig(_ context.Context, rc *rest.Config, token string, store *token.ReuseSourceStore) error { // nolint:gocyclo // mostly error handling
+	ex := rc.ExecProvider
+	if ex == nil {
+		return errors.New("an identity configuration was specified but the provided kubeconfig does not have execProvider section")
+	}
+	if ex.APIVersion != "client.authentication.k8s.io/v1" {
+		return errors.New("execProvider APIVersion is not client.authentication.k8s.io/v1")
+	}
+	if ex.Command != "up" || len(ex.Args) < 2 || ex.Args[0] != "organization" || ex.Args[1] != "token" {
```
luxas, line 32: there also is the "org" shorthand for organization :D

pkg/kube/client/upbound/upbound.go
```diff
@@ -0,0 +1,92 @@
...
+		return errors.New("execProvider command is not up organization token")
+	}
+
+	// Read the org name, it could either be the 3rd argument and provided with `ORGANIZATION` env var.
+	org := ""
+	if len(ex.Args) > 2 {
+		org = ex.Args[2]
+	}
+	for _, env := range ex.Env {
+		if env.Name == "ORGANIZATION" {
```
luxas, line 42: nit: make this a constant?

pkg/kube/client/upbound/upbound.go
```diff
@@ -0,0 +1,92 @@
...
+	})
+
+	return nil
+}
+
+// upboundTokenSource is an oauth2.TokenSource that fetches tokens from Upbound.
+type upboundTokenSource struct {
+	client       *auth.Client
+	org          string
+	refreshToken string
```
luxas, line 76: maybe it would be better to call this "staticToken" or "exchangeableToken", as at the moment this one is not updated?

we _should_ be using OAuth2 compliant refresh tokens instead soon (tm), that actually change on every refresh

pkg/kube/client/upbound/upbound.go
```diff
@@ -0,0 +1,92 @@
...
+	Base http.RoundTripper
+}
+
+func (s *upboundTokenSource) Token() (*oauth2.Token, error) {
+	resp, err := s.client.GetOrgScopedToken(context.Background(), s.org, s.refreshToken)
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot get upbound org scoped token")
+	}
+	return &oauth2.Token{
+		RefreshToken: s.refreshToken,
```
luxas, line 88: is this needed or does it work without? I would guess it works without, and if so I'd prefer to not expose the static token to callers

pkg/kube/client/upbound/upbound.go
```diff
@@ -0,0 +1,92 @@
...
+		client: auth.NewClient(&up.Config{
+			Client: up.NewClient(func(client *up.HTTPClient) {
+				client.BaseURL.Host = authHost
+			}),
+		}),
+		org:          org,
+		refreshToken: token,
+	}
+
+	rc.Wrap(func(rt http.RoundTripper) http.RoundTripper {
```
luxas, line 65: needed to check twice to understand that this indeed mutates the receiver. a godoc comment about that could be good

pkg/kube/client/upbound/upbound.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// upboundTokenSource is an oauth2.TokenSource that fetches tokens from Upbound.
+type upboundTokenSource struct {
+	client       *auth.Client
+	org          string
+	refreshToken string
+	// Base is the base RoundTripper used to make HTTP requests.
+	// If nil, http.DefaultTransport is used.
+	Base http.RoundTripper
```
luxas, line 79: unused?

pkg/kube/client/upbound/upbound.go
```diff
@@ -0,0 +1,92 @@
...
+		return errors.New("an identity configuration was specified but the provided kubeconfig does not have execProvider section")
+	}
+	if ex.APIVersion != "client.authentication.k8s.io/v1" {
+		return errors.New("execProvider APIVersion is not client.authentication.k8s.io/v1")
+	}
+	if ex.Command != "up" || len(ex.Args) < 2 || ex.Args[0] != "organization" || ex.Args[1] != "token" {
+		return errors.New("execProvider command is not up organization token")
+	}
+
+	// Read the org name, it could either be the 3rd argument and provided with `ORGANIZATION` env var.
```
luxas, line 36: _or_ provided?

pkg/kube/client/upbound/upbound.go
```diff
@@ -0,0 +1,92 @@
...
+	org := ""
+	if len(ex.Args) > 2 {
```
luxas, line 38: this does not error if we have more than 3, but I guess that's ok?
if we want, we can return error if len > 3

(or more nicely):
```
if !(len(ex.Args) == 2 || len(ex.Args) == 3) then error
```

pkg/kube/client/client.go
```diff
@@ -120,6 +143,21 @@ func configForProvider(ctx context.Context, local client.Client, providerConfigN
 					return nil, errors.Wrap(err, errInjectAzureCredentials)
 				}
 			}
+		case kconfig.IdentityTypeUpboundToken:
+			switch id.Source { //nolint:exhaustive
+			case xpv1.CredentialsSourceInjectedIdentity:
+				return nil, errors.Errorf("%s is not supported as identity source for identity type %s",
+					xpv1.CredentialsSourceInjectedIdentity, kconfig.IdentityTypeUpboundToken)
+			default:
+				tkn, err := resource.CommonCredentialExtractor(ctx, id.Source, b.local, id.CommonCredentialSelectors)
```
luxas, line 152: nit: static or exchangeable token

pkg/kube/client/upbound/upbound.go
```diff
@@ -0,0 +1,92 @@
...
+	client       *auth.Client
+	org          string
+	refreshToken string
+	// Base is the base RoundTripper used to make HTTP requests.
+	// If nil, http.DefaultTransport is used.
+	Base http.RoundTripper
+}
+
+func (s *upboundTokenSource) Token() (*oauth2.Token, error) {
+	resp, err := s.client.GetOrgScopedToken(context.Background(), s.org, s.refreshToken)
```
luxas, line 83: how long-lived / for what purpose is the context defined in calling WrapRESTConfig? shall we use that or no?

luxas (commented):
Thanks @turkenh, great PR 👏

pkg/kube/client/upbound/upbound.go
turkenh: Already relying on the fact that org is either the 3rd argument or provider with an env var. Erroring out otherwise, i.e. `organization name not provided in execProvider args or ORGANIZATION env var`. 

I don't see much value in being more strict here TBH, and would like to assume the command structure doesn't change (which would be breaking anyways).

pkg/kube/client/client.go
turkenh: Could you clarify, couldn't get.

pkg/kube/client/upbound/upbound.go
turkenh: hmm, it is the context set by controller-runtime reconciliation. 

I believe it makes sense to use it, this should be completed before that context is cancelled.

pkg/kube/client/upbound/upbound.go
turkenh: Renamed as `staticToken`.

examples/provider/provider-config-with-secret-upbound-identity.yaml
```diff
@@ -0,0 +1,18 @@
...
+  name: kubernetes-provider
+spec:
+  credentials:
+    source: Secret
+    secretRef:
+      namespace: crossplane-system
+      name: cluster-config
+      key: kubeconfig
+  identity:
+    type: UpboundToken
```
turkenh, line 13: @luxas any comment for this `identity.type`? Other supported values today are, `GoogleApplicationCredentials`, `AzureServicePrincipalCredentials` and `AzureWorkloadIdentityCredentials`. 

May be `UpboundTokens` (to be consistent with other with appended `s`)
Or `UpboundStaticTokens` if we expect to see tokens evolving in time as you pointed out in another comment.

turkenh: I didn't make any distinction between user tokens or robot tokens since the logic doesn't change.

pkg/kube/client/client.go
```diff
@@ -120,6 +143,21 @@ func configForProvider(ctx context.Context, local client.Client, providerConfigN
...
+				if err != nil {
+					return nil, errors.Wrap(err, errExtractUpboundCredentials)
+				}
+
+				if err := upbound.WrapRESTConfig(ctx, rc, string(tkn), b.store); err != nil {
```
turkenh, line 157: TODO: Trim whitespaces around the token. 

Thanks a lot @haarchri, for testing the PR and catching this 🙌

haarchri:
@turkenh thanks for the implementation - i tested the implementation over time i got the following issue:

```
  Warning  CannotObserveExternalResource  2m8s (x2110 over 34h)  managed/object.kubernetes.crossplane.io  cannot get object: failed to get API group resources: unable to retrieve the complete list of server APIs: spaces.upbound.io/v1beta1: Get "https://upbound-gcp-us-west-1.space.mxe.upbound.io/apis/spaces.upbound.io/v1beta1": cannot get upbound org scoped token: Post "https://auth.upbound.io/apis/tokenexchange.upbound.io/v1alpha1/orgscopedtokens": context canceled
```

haarchri:
yes looks like after 60 minutes something happens:

```
Events:
  Type     Reason                         Age                   From                                     Message
  ----     ------                         ----                  ----                                     -------
  Warning  CannotObserveExternalResource  2m23s (x46 over 42m)  managed/object.kubernetes.crossplane.io  cannot get object: failed to get API group resources: unable to retrieve the complete list of server APIs: spaces.upbound.io/v1beta1: Get "https://upbound-gcp-us-west-1.space.mxe.upbound.io/apis/spaces.upbound.io/v1beta1": cannot get upbound org scoped token: Post "https://auth.upbound.io/apis/tokenexchange.upbound.io/v1alpha1/orgscopedtokens": context canceled
```



```
NAME                                            KIND           PROVIDERCONFIG                               SYNCED   READY   AGE
object.kubernetes.crossplane.io/ctp-dev-69wbr   Secret         default                                      False    True    61m
object.kubernetes.crossplane.io/ctp-dev-bwqbl   ControlPlane   5ab0024e-fbc3-4a42-a1a1-a1dd4875093d-space   False    True    61m
object.kubernetes.crossplane.io/ctp-dev-qcj2n   Secret         default                                      False    True    61m
```

examples/provider/provider-config-with-secret-upbound-identity.yaml
turkenh: Renamed as `UpboundToken` -> `UpboundTokens` for now.

pkg/kube/client/upbound/upbound.go
turkenh: But, it caused problems when the token expires, and we need to do another refresh: https://github.com/crossplane-contrib/provider-kubernetes/pull/251#issuecomment-2132159590

This is expected since the context from the initial call becomes cancelled during refresh. I couldn't find a clean way without duplicating a bunch of code from the oauth2 library, hence used a new context with a deadline.

pkg/kube/config/config.go
```diff
@@ -0,0 +1,66 @@
...
+
+// Package config contains API types used by Crossplane providers interacting
+// with Kubernetes APIs.
+// +kubebuilder:object:generate=true
+package config
+
+import xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+
+// IdentityType used to authenticate to the Kubernetes API.
+// +kubebuilder:validation:Enum=GoogleApplicationCredentials;AzureServicePrincipalCredentials;AzureWorkloadIdentityCredentials;UpboundToken
```
haarchri, line 25: @turkenh think here is bug 
```suggestion
// +kubebuilder:validation:Enum=GoogleApplicationCredentials;AzureServicePrincipalCredentials;AzureWorkloadIdentityCredentials;UpboundTokens
```

turkenh: 🤦 fixed!

haarchri:
its working now:
```
object.kubernetes.crossplane.io/ctp-prod-cdd6x       Secret         default                                      True     True    83m
object.kubernetes.crossplane.io/ctp-prod-lgbtg       Secret         default                                      True     True    83m
object.kubernetes.crossplane.io/ctp-prod-wfzd4       ControlPlane   24bf1d49-140c-41e2-819e-15e14acf2538-space   True     True    83m
```

thanks for the implementation

pkg/kube/client/client.go
luxas: sorry: tried to suggest that the variable being returned `tkn` could be called `staticToken` or `exchangeableToken` to it's clear this is the "before" state.

pkg/kube/client/client.go
```diff
@@ -120,6 +144,21 @@ func configForProvider(ctx context.Context, local client.Client, providerConfigN
...
+			case xpv1.CredentialsSourceInjectedIdentity:
+				return nil, errors.Errorf("%s is not supported as identity source for identity type %s",
+					xpv1.CredentialsSourceInjectedIdentity, kconfig.IdentityTypeUpboundTokens)
+			default:
+				tkn, err := resource.CommonCredentialExtractor(ctx, id.Source, b.local, id.CommonCredentialSelectors)
+				if err != nil {
+					return nil, errors.Wrap(err, errExtractUpboundCredentials)
+				}
+
+				if err := upbound.WrapRESTConfig(ctx, rc, strings.TrimSpace(string(tkn)), b.store); err != nil {
```
luxas, line 158: if this was a problem before, maybe add a comment that trimspace actually is needed and why

luxas (approved):
Thanks @turkenh!

lsviben (approved):
LGTM, thanks @turkenh!

PR #274, opened 2024-07-23
Fix poll jittering in Object controller

aerfio:

This PR fixes Object controller which sometimes returns a negative poll interval if the reconciled Object instance is not ready, which is then ignored by controller-runtime internals.

In my case this has been observed in a situation in which the k8s object from `spec.forProvider.manifest` takes a while to get ready, but under 10m which is the default `pollInterval`. 

The logic which calculates the pollInterval does not take into account that the base `pollInterval` has been updated to 30 seconds if the reconciled resource is not ready. It took the base value of `pollJitter` which is 1 minute, and calculated next poll interval using this formula:
```go
return pollInterval + time.Duration((rand.Float64()-0.5)*2*float64(pollJitter))
```
I guess that most users run provider-kubernetes without adjusting [default poll interval](https://github.com/crossplane-contrib/provider-kubernetes/blob/a9c6ef8b02f8f56a144d9103bddcab2e8a04b3f6/cmd/provider/main.go#L60), so if my math is right if `rand.Float64` returns anything less then `0.25` (its documentations states: "// Float64 returns, as a float64, a pseudo-random number in the half-open interval [0.0,1.0)"), the returned value is less then 0. 25% chance to run into this bug. And if I understand the c-r source code right, [it ignores negative RequeueAfter](https://github.com/kubernetes-sigs/controller-runtime/blob/1ed345090869edc4bd94fe220386cb7fa5df745f/pkg/internal/controller/controller.go#L312).

The solution is to recalculate the pollInterval taking into account the `pollJitterPercentage` - this means that with default value of poll jitter precentage the poll interval would be in half-open interval [27.0, 33.0) (seconds).

I have:

### DIFF
--- a/cmd/provider/main.go
+++ b/cmd/provider/main.go
@@ -166,7 +166,7 @@ func main() {
 	// notice and remove when we drop support for v1alpha1.
 	kingpin.FatalIfError(ctrl.NewWebhookManagedBy(mgr).For(&v1alpha1.Object{}).Complete(), "Cannot create Object webhook")
 
-	kingpin.FatalIfError(object.Setup(mgr, o, *sanitizeSecrets, pollJitter), "Cannot setup controller")
+	kingpin.FatalIfError(object.Setup(mgr, o, *sanitizeSecrets, pollJitter, *pollJitterPercentage), "Cannot setup controller")
 	kingpin.FatalIfError(mgr.Start(ctrl.SetupSignalHandler()), "Cannot start controller manager")
 }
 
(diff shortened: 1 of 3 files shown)


turkenh (approved):
Great catch 🙌  and awesome PR description 🙇‍♂️ Thanks a lot!

aerfio:
@turkenh Hi, can I count on a patch release with this fix? 🙏🏻

turkenh:
@aerfio sure, https://github.com/crossplane-contrib/provider-kubernetes/releases/tag/v0.14.1 is out.

aerfio:
@turkenh Thank you!

PR #301, opened 2024-10-11
feat: make the webhook port configurable

### LINKED ISSUE #208: Conversion Webhook breaks deployment on clusters with seperate control plane
### What happened?
After upgrading from 0.10.0 to 0.12.0 the provider is unable to start up successfully with logs indicating the CRDs cannot be watched/listed:

```
W0226 16:25:45.849143       1 reflector.go:539] k8s.io/client-go@v0.29.1/tools/cache/reflector.go:229: failed to list *v1alpha2.Object: conversion webhook for kubernetes.crossplane.io/v1alpha1, Kind=Object failed: Post "https://provider-kubernetes.crossplane-system.svc:9443/convert?timeout=30s": Address is not allowed
E0226 16:25:45.849254       1 reflector.go:147] k8s.io/client-go@v0.29.1/tools/cache/reflector.go:229: Failed to watch *v1alpha2.Object: failed to list *v1alpha2.Object: conversion webhook for kubernetes.crossplane.io/v1alpha1, Kind=Object failed: Post "https://provider-kubernetes.crossplane-system.svc:9443/convert?timeout=30s": Address is not allowed
crossplane-kubernetes-provider: error: Cannot start controller manager: failed to wait for providerconfig/providerconfig.kubernetes.crossplane.io caches to sync: timed out waiting for cache to be synced for Kind *v1alpha1.ProviderConfig
Stream closed EOF for crossplane-system/provider-kubernetes-6e1fd76ec9c7-65d577488d-jzgzs (package-runtime)
```

Our Kubernetes cluster is an AWS EKS with Calico as the CNI, therefore the cluster control plane and pods running in the cluster are running in different networks (as the CNI cannot be applied to the AWS managed control plane). The consequence of this is that every webhook needs to be run with `hostNetwork: true`. 

As the conversion from v1alpha1 to v1alpha2 is fairly straight forward and could be done manually instead, maybe a command line option can be introduced to disable the webhook. Although this also affects the CRDs and thus depends on the code generation used by crossplane providers, but I might be wrong here.

### How can we reproduce it?

Deploy the provider in version 1.11+ and block network access from the control plane to the webhook.

Crossplane version: v1.15.0
Cloud Provider: AWS
Distribution: EKS v1.29
Container Network Interface: Calico

Argannor:

As outlined in the issue below, when running Crossplane in certain environments (like EKS using Calico CNI), where the control plane resides outside the workload network it is necessary to run webhooks in the host network. This calls for the need to make the port configurable to avoid conflicts.

I opened [crossplane/crossplane!5540](https://github.com/crossplane/crossplane/pull/5540) to allow a `DeploymentRuntimeConfig` to overwrite the ports of the crossplane managed services created for each provider / function.

Fixes #208

I have:

I deployed my fork of this and crossplane in our test cluster running EKS and Calico. With this I was able to run both Crossplane and provider-kubernetes inside the host network with custom ports.

This change only makes sense if my other PR gets accepted.

### DIFF
--- a/cmd/provider/main.go
+++ b/cmd/provider/main.go
@@ -35,6 +35,7 @@ import (
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
@@ -257,7 +258,22 @@ func main() {
 		kingpin.FatalIfError(controllerCluster.Setup(mgr, o, po), "Cannot setup cluster-scoped controller")
 		kingpin.FatalIfError(controllerNamespaced.Setup(mgr, o, po), "Cannot setup namespaced controller")
 	}
+
+	// Setup health probes
+	kingpin.FatalIfError(setupHealthProbes(mgr), "Cannot setup health probes")
+
 	kingpin.FatalIfError(mgr.Start(ctrl.SetupSignalHandler()), "Cannot start controller manager")
+}
+
+// setupHealthProbes sets up the health and readiness probes.
+func setupHealthProbes(mgr ctrl.Manager) error {
+	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+		return errors.Wrap(err, "cannot add readiness probe")
+	}
+	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
+		return errors.Wrap(err, "cannot add liveness probe")
+	}
+	return nil
 }
 
 // UseISO8601 sets the logger to use ISO8601 timestamp format
--- a/cmd/provider/main.go
+++ b/cmd/provider/main.go
@@ -38,6 +38,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
+	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
 	changelogsv1alpha1 "github.com/crossplane/crossplane-runtime/v2/apis/changelogs/proto/v1alpha1"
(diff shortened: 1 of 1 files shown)


cmd/provider/main.go
```diff
@@ -63,6 +63,7 @@ func main() {
 		leaderElection          = app.Flag("leader-election", "Use leader election for the controller manager.").Short('l').Default("false").Envar("LEADER_ELECTION").Bool()
 		maxReconcileRate        = app.Flag("max-reconcile-rate", "The number of concurrent reconciliations that may be running at one time.").Default("100").Int()
 		sanitizeSecrets         = app.Flag("sanitize-secrets", "when enabled, redacts Secret data from Object status").Default("false").Envar("SANITIZE_SECRETS").Bool()
+		webhookPort             = app.Flag("webhook-port", "The port the webhook listens on").Default("9443").Envar("WEBHOOK_PORT").Int()
```
ravilr, line 66: for all providers, also allow configuring metrics bind address which gets passed onto sigs.k8s.io/controller-runtime/pkg/metrics/server.Options ?
we can standardize on `--webhook-port` and `--metrics-bind-address`  flags across all providers, similar to crossplane core flags..

Argannor: +1 for standardizing. I added the --metrics-bind-address as well as requested. Thank you for the feedback!

ctenberge-oneid:
I could really use this sooner rather than later. @ravilr is there anyway this can be prioritized? Really appreciate your work either way.

ctenberge-oneid:
What is stopping this from being merged? Thanks for the work by the way.

erhancagirici (commented):
LGTM in general. nit would be whether it would be feasible to enforce standardization of those flag names consistently across all providers. That can be a follow-up though.

jbw976 (dismissed):
>  feasible to enforce standardization

enforcement, probably not across the entire ecosystem i suppose, but let's try to keep them standard for all the providers we can, including this one 😉 

see https://github.com/crossplane-contrib/provider-kubernetes/pull/301/files#r2077361313

cmd/provider/main.go
ravilr: please see https://github.com/crossplane/crossplane/pull/5540#discussion_r1823349314

the argument towards keeping the metrics port configuration as address string is: Some Security focussed setups **would not** want to expose the metrics port on all interfaces, instead only on 127.0.0.1:8080 and then front it with a [rbac-proxy](https://github.com/brancz/kube-rbac-proxy) sidecar which requires auth for the prometheus scrapers to be able to reach it.

Intaking it as `metrics-port` int arg, will take away above ability.

jonasz-lasut:
Thank you for your contribution @Argannor I've rebased the PR and standardized the flag names and defaults to be compliant with the rest of providers (provider-helm & upjet-based providers).

@jbw976 FYI we've discussed this in the ecosystem team and we'd rather keep the address flag which is different than the one in core in both name and functionality as it gives more flexibility and was already standardized in upjet-based providers

jonasz-lasut (approved):

PR #329, opened 2025-01-06
Add DesiredStateChanged event filter to filter out status updates

ravilr:

- Added an event filter using the [DesiredStateChanged](https://github.com/crossplane/crossplane-runtime/blob/0eae57e9c06bb807864bda7c1eece43671698ab7/pkg/resource/predicates.go#L51) predicate to the resource controllers so that updates to only labels, annotations (Crossplane's create pending/failed annotations will be ignored) and the spec will result in reconcile requests.

I have:

- make test

- without this change,  when a `Object` resource's resourceVersion keeps changing due to changing `metadata.managedFields` of the managed K8s resource in `status.atProvider`, results in repeated reconciles of such Object in a short period of time( teal colored peaks in the below metrics graph).  With this change, such status updates doesn't result in reconcile spins (green plot, after the change):

Screenshot 2025-01-06 at 1 11 39 PM

### DIFF
--- a/internal/controller/object/object.go
+++ b/internal/controller/object/object.go
@@ -191,6 +191,7 @@ func Setup(mgr ctrl.Manager, o controller.Options, sanitizeSecrets bool, pollJit
 
 	cb := ctrl.NewControllerManagedBy(mgr).
 		Named(name).
+		WithEventFilter(resource.DesiredStateChanged()).
 		WithOptions(o.ForControllerRuntime()).
 		For(&v1alpha2.Object{})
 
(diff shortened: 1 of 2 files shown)


ravilr:
@bobh66  PTAL.  we observe considerable reconciles spinning behavior leading to K8s client qps peaks, without filtering out the status. 

 Do you have more details on what you observed in https://github.com/crossplane-contrib/provider-kubernetes/pull/110#issuecomment-1496165241 with the event filter enabled ?

bobh66:
LGTM

turkenh (approved):
Thanks @ravilr, looking good to me as well.

PR #349, opened 2025-03-06
Enable deleting objects when identity of manifest changes

### LINKED ISSUE #287: Changing Object.spec.forProvider.manifest.metadata.name orphans previous resource
### What happened?
After changing the Object.spec.forProvider.manifest.metadata.name the resource with the previous name does not get deleted.

### How can we reproduce it?
Create a Resource:
```yaml
kind: Object
metadata:
  name: foo
spec:
  forProvider:
    manifest:
      apiVersion: v1
      kind: ConfigMap
      metadata:
        namespace: default
        name: test1
```
Change the Resource
```yaml
kind: Object
metadata:
  name: foo
spec:
  forProvider:
    manifest:
      apiVersion: v1
      kind: ConfigMap
      metadata:
        namespace: default
        name: test2
```

run `kubectl get cm -n default` and observe both test1 and test2 are present.

Provider Kubernetes: v0.14.1
Crossplane version: v1.16.0
Kubernetes: v1.28.8

nkvoll:

This is an attempt to move the discussion of https://github.com/crossplane-contrib/provider-kubernetes/issues/287 forward.

I think some of the issue is that the interfaces provided by `crossplane-runtime` (e.g the `managed.ExternalObservation`) doesn't allow to neatly encode the `Observe` function wanting to return a the operations required for a rename (e.g a Delete + Create). Without changing interfaces etc, which may be a heavy ask for the needs of potentially a single provider, we might have to be a little creative.

At first, I tried dealing with the Deletion of the previous resource in `Create`, but its problematic because `Observe` will try to update `atProvider`, which overwrites the information we need to delete a previously managed object.

Doing the deletion in `Observe` is a less noisy option code-wise, but of course means that "Observing" might entail "Delete something", and I'd love to hear opinions on this with possible solutions (e.g like updating the interface/structs from the runtime, expanding the Object resource to include resource references etc -- these are larger lifts, but it would be great to have clarity whether doing the requisite work for that would end up with the work being merged before committing to it).

Fixes #287

I have:

Running locally in a local dev environment.

### DIFF
--- a/internal/controller/object/object.go
+++ b/internal/controller/object/object.go
@@ -89,6 +89,7 @@ const (
 	errGetObservedState        = "cannot get observed state"
 	errGetDesiredState         = "cannot get desired state"
 	errUnmarshalTemplate       = "cannot unmarshal template"
+	errUnmarshalAtProvider     = "cannot unmarshal atProvider"
 	errFailedToMarshalExisting = "cannot marshal existing resource"
 
 	errGetReferencedResource       = "cannot get referenced resource"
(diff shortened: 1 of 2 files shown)


bobh66:
I'm not convinced that we want the provider to delete an existing resource and replace it with another one.  We explicitly _don't_ do that in the terraform-based providers when it's necessary to delete a resource to apply an update.  I understand the scenarios involved, but if kubernetes requires a delete/create then I think provider-kubernetes should also require a delete/create, which can be handled at the Composition layer by changing the composite resource name.

It seems a little too much like magic and could have unintended consequences if someone doesn't understand that changing the name will cause the resource to be deleted and recreated.

It makes me wonder whether there should be CEL rules in the Object CRD to prevent changes to `spec.manifest.apiVersion`, `spec.manifest.kind`, `spec.manifest.metadata.name` and `spec.manifest.metadata.namespace` since all of these would trigger a delete/create if changed.

nkvoll:
Preventing these changes would also solve my concern (inadvertently changing the identity, potentially leaving orphans behind without realizing it). 👍

jonasz-lasut:
Closing as the issue that was linked to this PR is also closed, this is intended behavior and I agree with Bob that we can think about adding CEL rules to ensure immutability of `apiVersion`, `kind`, `metadata.name` and `metadata.namespace` in the `spec.manifest`

PR #350, opened 2025-03-07
Fixes an issue when enable-server-side-apply and sanitize-secrets are turned on

ravilr:

- Fixes an issue seen when `--enable-server-side-apply` and `--sanitize-secrets` are turned on in the provider-kubernetes.

With both those flags enabled, an `Object` resource managing any Secret resource never reaches `Ready` . the redaction of secret data in the Object status was mutating the current Observed object, which is being used to compare with the desired state in equality check for successful syncing. This was not an issue with client-side patch applicator, as it uses the data in the `last-applied-configuration` annotation for sync comparison check.

Tested with:

- `kubectl apply  -f object.yaml`  never reaches Ready with the provider started with args `--sanitize-secrets` and `--enable-server-side-apply`:

```
kind: Object
metadata:
  name: foo
spec:
  references:
  - patchesFrom:
      apiVersion: v1
      kind: Secret
      name: bar
      namespace: default
      fieldPath: data.key
    toFieldPath: data.key-from-bar
  forProvider:
    manifest:
      apiVersion: v1
      kind: Secret
      metadata:
        namespace: default
  providerConfigRef:
    name: kubernetes-provider # replace/create providerconfig
---
apiVersion: v1
kind: Secret
metadata:
  name: bar
  namespace: default
stringData:
  key: some-value
```

@turkenh  PTAL 🙏 

I have:

`make test` and manually tested applying above manifest.

### DIFF
--- a/internal/controller/object/object.go
+++ b/internal/controller/object/object.go
@@ -482,16 +482,18 @@ func parseManifest(obj *v1alpha2.Object) (*unstructured.Unstructured, error) {
 func (c *external) setAtProvider(obj *v1alpha2.Object, observed *unstructured.Unstructured) error {
 	var err error
 
+	// sanitize/mutate only the copied object
+	sObserved := observed.DeepCopy()
 	if c.sanitizeSecrets {
 		if observed.GetKind() == "Secret" && observed.GetAPIVersion() == "v1" {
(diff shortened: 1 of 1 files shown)


turkenh (approved):
Great, thanks for the fix 🙌

PR #368, opened 2025-06-17
build: pass robot token credentials to publishing to mirror registry

jbw976:

Now that https://github.com/crossplane-contrib/provider-workflows/pull/15 has been merged, this PR updates the publishing workflow to pass the correct robot credentials. It looks like this repo already had some defined, so let's reuse them 🙏 

Once this PR is merged, we should still successfully continue to publish to ghcr.io and mirror to xpkg.upbound.io.

I have:

It hasn't explicitly been tested - we'll see how CI goes, but I expect the real path won't be tested until we run a publishing 😇 


jbw976:
I was planning on re-running the last publishing basically @chlunde and verifying it still worked OK, had access to the registries, etc.

Issue #259, opened 2024-05-24
Remove finalizer based deletion ordering functionality


turkenh:
### What problem are you facing?

Provider Kubernetes introduced a [finalizer based](https://github.com/crossplane-contrib/provider-kubernetes/blob/main/docs/enhanced-provider-k8s.md#using-finalizers) deletion ordering solution before Crossplane introduces [Usages](https://github.com/crossplane/crossplane/blob/master/design/one-pager-generic-usage-type.md). The functionality works fine most of the time, however, may cause ending up with stuck resources with finalizers not removed especially under heavy conditions where multiple Objects deleted in parallel or provider gets restarted (just theory, haven't dug deeper). 

We should deprecate this stop-gap solution in favor of Crossplane Usages to avoid the maintenance burden and confusion with multiple solutions to the same problem.

### How could Crossplane help solve your problem?

Deprecate & remove finalizer based deletion ordering functionality. //cc @morningspace

rperez95:
Hi @turkenh,

I understand the reasoning behind deprecating the finalizer-based deletion ordering functionality, but we're concerned about how it might affect certain use cases.

For instance, we rely on this feature to prevent deletion of pre-created k8s secrets used by some of our compositions. As far as we know, the use of `Usages` is restricted to Crossplane resources, which does not include k8s secrets.

We use `Usages` to deletion ordering, but we also use the finalizer-based functionality to ensure that k8s secrets are not deleted before the managed resources deployed by our compositions are fully deleted.

Could you clarify whether an alternative mechanism would be introduced to handle such scenarios, or provide guidance on how to address this gap if the finalizer-based approach is removed?

turkenh:
> Hi [@turkenh](https://github.com/turkenh),
> 
> I understand the reasoning behind deprecating the finalizer-based deletion ordering functionality, but we're concerned about how it might affect certain use cases.
> 
> For instance, we rely on this feature to prevent deletion of pre-created k8s secrets used by some of our compositions. As far as we know, the use of `Usages` is restricted to Crossplane resources, which does not include k8s secrets.
> 
> We use `Usages` to deletion ordering, but we also use the finalizer-based functionality to ensure that k8s secrets are not deleted before the managed resources deployed by our compositions are fully deleted.
> 
> Could you clarify whether an alternative mechanism would be introduced to handle such scenarios, or provide guidance on how to address this gap if the finalizer-based approach is removed?

[...]

PR #372, opened 2025-06-23
fix: Retry adding finalizers to referenced objects on conflict

### LINKED ISSUE #371: Race Condition in Parallel Reconciliation Prevents Finalizer Addition to Referenced Objects
### What happened?
In the provider-kubernetes, a race condition during parallel resource reconciliation prevents finalizers from being reliably added to referenced objects. This can lead to orphaned resources upon deletion and an inconsistent state.

The expected behavior is that all necessary finalizers are added to all referenced objects, even if initial attempts conflict.

The actual behavior is that the reconciliation logic has a flaw that causes it to fail silently. The flow is as follows:

* The provider starts reconciling multiple resources that reference the same object at the same time.
* During reconciliation, a resource first adds a finalizer to itself. This step succeeds.
* Immediately after, it attempts to add its finalizer to the shared referenced objects. Due to the parallel reconciliations, this frequently causes a conflict, and the reconciliation request is correctly re-queued.
* On the next reconciliation attempt, the controller sees that the resource's own finalizer has already been set and returns early.
* This early return means the controller never retries adding the finalizer to the referenced objects that it failed to update in the first attempt.

As a result, only the resources that won the initial race add their finalizers to the referenced objects. The others will never have their finalizers added, breaking the garbage collection mechanism.

### How can we reproduce it?
* Define multiple Object resources in provider-kubernetes where several objects contains references (`dependsOn`) to a single object.
* Arrange for these Object resources to be created and reconciled.
* Inspect the referenced object.
* Observe the metadata.finalizers array on the object. You will likely see that finalizers from some of the Object resources are missing, indicating that they lost the race and will never be added due to the bug.

Crossplane version: 1.19.1
provider-kubernetes version: v0.18.0
Kubernetes version: Client Version: v1.33.1, Kustomize Version: v5.6.0 Server Version: v1.33.1
Kubernetes distribution: Kind

helderjs:

This pull request updates the finalizer logic for Object resources and their references in the provider-kubernetes project. It ensures that finalizers are correctly added to both the managed resource and any referenced resources, improving cleanup and dependency handling.

Fixes #371 

I have:

This change has been covered by comprehensive unit tests in internal/controller/object/object_test.go,


turkenh (changes requested):
Thanks for the PR @helderjs. 

However, this is a functionality that we plan to deprecate, so, I am not willing to introduce behavioral changes at this moment. See https://github.com/crossplane-contrib/provider-kubernetes/issues/259

PR #378, opened 2025-07-23
Add support to impersonate SA when GoogleApplicationCredentials identity type is used

kostyaplis:

**Use case**: 
Multiple GKE environments across different GCP projects are managed by Crossplane. Each environment has a preconfigured `management` service account, which can be impersonated by the Crossplane service account. 
After provisioning GKE clusters, we want to manage Kubernetes API objects using `provider-kubernetes`, leveraging the same authentication flow—specifically, impersonating the respective service account in each target project.

This pull request adds support for setting `identity.impersonatedServiceAccount.name`when `identity.type: GoogleApplicationCredentials` is used. This allows the retrieved credentials to be exchanged for an impersonated service account token.

I have:

This code has been built, published to a private repo, and tested with Crossplane v1.20 in our GCP environment. As we don't see the way to test google auth in local environment.

For reference, the built image has been pushed to Docker Hub: https://hub.docker.com/r/kostyaplis/provider-kubernetes/tags

Please let us know if the logic introduced aligns with the provider's design principles, so this feature has the potential to be officially released.
If not, we would appreciate any guidance on whether the same authentication flow can be achieved using existing tools. Thank you in advance!


kostyaplis:
Hello maintainers!

Any feedback on this PR is highly appreciated. Thanks

kostyaplis:
Superseded by https://github.com/crossplane-contrib/provider-kubernetes/pull/528

PR #387, opened 2025-08-27
Feature: use proxy-url from cluster in kubeconfig if set

shepses:

Clusters in kubeconfig may have a field `proxy-url` that needs to be respected to reach the specified endpoints.
Currently, `client.go` in the provider does not respect this.

`struct rest.Config` created in `client.go` offers a field `proxy:func(*http.Request) (*url.URL, error)` that may be used to this end. 

Just compile/deploy-local as of now. Want to see if there is interest.

### DIFF
--- a/pkg/kube/client/client.go
+++ b/pkg/kube/client/client.go
@@ -16,6 +16,8 @@ package client
 import (
 	"context"
 	"strings"
+	"net/http"
+	"net/url"
 
 	"github.com/pkg/errors"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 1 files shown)


mcfarlanem:
@veith4f @turkenf it's disappointing we haven't got this contribution merged yet.  Since my last post I have literally deployed hundreds of clusters using this.  IMHO this is a better solution than using env vars at the pod level, not to mention it taking advantage of core Kubernetes capability.

veith4f:
How bout it guys. Can we get this merged?

PR #416, opened 2025-12-05
feat: promote Server-side apply to beta - enabled by default

erhancagirici:

⚠️ Promote server-side apply feature to `beta` and enable by default

`--enable-server-side-apply` flag now defaults to `true`. 
To disable server-side apply pass the negated flag `--no-enable-server-side-apply`

Added handling for migrating managedField entries to SSA manager. This is to address the case at https://github.com/kubernetes/kubernetes/issues/99003 issue description.

Added E2E tests for SSA and SSA field manager upgrade.

I have:

Added new e2e tests

### DIFF
--- a/docs/server-side-apply.md
+++ b/docs/server-side-apply.md
@@ -1,0 +1,59 @@
+# Server-side apply in provider-kubernetes
+
+For managing and syncing k8s resources, provider-kubernetes uses server-side apply (SSA)
+by default.
+See [official kubernetes documentation](https://kubernetes.io/docs/reference/using-api/server-side-apply/)
+for reference on how SSA works in Kubernetes.
+
+### General behavior of SSA
+
+The `Object` managed resources (MRs) in `kubernetes.crossplane.io/v1alpha1`(cluster-scoped)
+and `kubernetes.m.crossplane.io/v1alpha1`(namespace-scoped) set the fully specified intent
+for the desired k8s object in their `.spec.forProvider.manifest`.  
+
+For a given `Object` MR, `provider-kubernetes` uses the field manager `provider-kubernetes/object-mr-name`
+while applying the desired k8s resource with SSA. The provider syncs and detect drifts only for the fields
+it manages, that are specified in the `.spec.forProvider.manifest` in accordance with K8s SSA mechanics.
+
+Per [k8s recommendation for controllers](https://kubernetes.io/docs/reference/using-api/server-side-apply/#using-server-side-apply-in-a-controller)
+,in case of conflicts on field management `provider-kubernetes` forces the conflicts.
+
+> [!WARN]
+> Care should be taken if there are other external controllers managing the same k8s resource.
+> They should not manage the same fields, as they might race on the field ownerships.
+
+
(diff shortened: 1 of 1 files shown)


internal/controller/namespaced/object/syncer.go
```diff
@@ -4,9 +4,13 @@ import (
 	"context"
```
turkenh: Not introduced with this PR, but how do we manage keeping `cluster/object/` and `namespaced/object/` in sync? IIUC, this is a manual process right? Wondering if we could automate it somehow, e.g. change in one place and make generate distibutes.

examples/namespaced/object/testhooks/disable-ssa.sh
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env bash
+set -aeuo pipefail
+
+echo " ⏳ Disabling SSA feature for the provider..."
+${KUBECTL} patch deploymentruntimeconfig runtimeconfig-provider-kubernetes --type='json' -p='[{"op":"replace","path":"/spec/deploymentTemplate/spec/template/spec/containers/0/args", "value":["--debug", "--no-enable-server-side-apply"]}]'
```
turkenh: > --no-enable-server-side-apply

Oh, interesting, can we set this flag to false like this? vs `--enable-server-side-apply=false`

internal/controller/namespaced/object/syncer.go
```diff
@@ -110,8 +126,69 @@ func (s *SSAResourceSyncer) GetDesiredState(ctx context.Context, obj *v1alpha1.O
...
+	if current == nil || !s.needSSAFieldManagerUpgrade(current) {
+		return nil
+	}
+	// this generates a raw json patch that transfers only the specified
+	// CSA field managers to the SSA field manager. Other field managers
+	// that might exist are not modified.
+	// The generated patch includes resourceVersion, i.e. uses optimistic
+	// locking, so if the object was changed between the last observation
+	// and the patch, it will be safely rejected, then requeued.
+	mfUpgradePatch, err := csaupgrade.UpgradeManagedFieldsPatch(current, s.legacyCSAFieldManagers, ssaFieldOwner(obj.GetName()))
```
turkenh: This function [may return](https://github.com/kubernetes/client-go/blob/54601aa99ffde1d9a79128043b852f41fad21415/util/csaupgrade/upgrade.go#L120) a `nil` patch if nothing needs to be done. So, it is a good idea to handle nil patches, e.g. return nil and not call client.patch.

internal/controller/namespaced/object/syncer.go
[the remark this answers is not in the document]
erhancagirici: yes, fixed it 👍

internal/controller/namespaced/object/syncer.go
erhancagirici: Yes, this is a manual process at the moment. I'll create an issue for this 👍 
IMHO it needs some refactor first. Automation was not trivial, as there were some subtle differences in namespaced vs cluster-scoped APIs.

PR #417, opened 2025-12-09
ci: bump golangci-lint to v2 and GH action dependencies

erhancagirici:

Upgrade golangci-lint to v2
- migrate `.golangci.yml`
- fix linter issues after bump

Bump go version to 1.24.11

Bumps all GH action dependencies

I have:

CI passes

note: codecov is reporting a false-positive due to comparison with wrong base commit. These are due to missing coverage reports for latest main, failing to upload reports. After merging this PR it will be fixed.


.golangci.yml
phisco: nit: would be nice to preserve comments, where applicable, I did that pretty effectively with claude code in the past.

erhancagirici: done 👍

PR #429, opened 2026-03-05
Add ThreadSafe client caching to IdentityAwareBuilder

### LINKED ISSUE #430: [Perf][IdentityAwareBuilder] Client connection pooling for ProviderConfig identities
## Problem
Currently, IdentityAwareBuilder.KubeForProviderConfig() creates a new Kubernetes client on every reconcile, even when the same ProviderConfig (and identity) was used previously. 
This causes:
Connection churn – repeated TCP/TLS handshakes to the same remote cluster
Memory pressure – short-lived clients accumulate before GC
Token refresh overhead – OAuth tokens are re-fetched 

## Solution
Add a thread-safe client cache to IdentityAwareBuilder with TTL-based expiration:

Cache key generation – derive a stable key from the ProviderConfig's identity source 
Cache lookup – before creating a client, check for a valid (non-expired) cached entry
TTL expiration – cached clients expire after 30 minutes (configurable) to ensure credential rotation is respected
Lazy cleanup – expired entries are evicted on next access rather than requiring a background goroutine

siddimore:
Adds thread-safe client caching to IdentityAwareBuilder to avoid creating redundant Kubernetes clients for the same ProviderConfig identity. This reduces connection overhead when multiple Objects reference the same ProviderConfig.

**Changes**:
- Added cachedClient struct wrapping client.Client with createdAt timestamp
- Added cache (sync.Map) and cacheTTL fields to IdentityAwareBuilder
- KubeForProviderConfig() now checks cache before creating clients
- Added cacheKeyForProviderConfig() helper generating stable keys from identity source
- Default TTL: 30 minutes (change to configurable)

Fixes #430 

I have:

Defined new tests:
- Builder construction
- Cache key generation 
- Cache key uniqueness
- Client expiry behavior
- Cache hit with valid client

### DIFF
--- a/pkg/kube/client/client.go
+++ b/pkg/kube/client/client.go
@@ -15,7 +15,10 @@ package client
 
 import (
 	"context"
+	"fmt"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/pkg/errors"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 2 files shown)


jonasz-lasut:
Superseded by https://github.com/crossplane-contrib/provider-kubernetes/pull/543 && https://github.com/crossplane-contrib/provider-kubernetes/pull/560

PR #437, opened 2026-03-23
Add configurable Object DeletionPropagationPolicy

### LINKED ISSUE #342: Deleting an `Object` which created a `Job` does not delete the pods created by that job
### What happened?

I have a Composition that creates an `Object` which creates a `Job`. When the `Object` is deleted the `Job` is deleted but the `Pods` which were created to execute the `Job` are not deleted. The expected behavior is that deleting the `Object` will also delete the `Pods` since this is what happens when a `Job` is deleted manually. This does not happen when using a `Deployment` instead of a `Job`, in that case `Pods` are deleted when the `Deployment` is deleted by way of deleting the `Object`

### How can we reproduce it?

1. Create this `Object`

This job will exit with a random code between 0 and 5 to simulate a job which needs to retry a few times which is the case for the real job in my Composition.

```yaml
kind: Object
metadata:
  name: test-job
spec:
  references:
  forProvider:
    manifest:
      apiVersion: batch/v1
      kind: Job
      metadata:
        name: test-job
        namespace: default
      spec:
        template:
          spec:
            containers:
              - name: busybox
                image: busybox
                command: ["/bin/sh", "-c", "exit $((RANDOM % 6))"]
                imagePullPolicy: IfNotPresent
            restartPolicy: Never
```

2. Allow it to create a few pods. It does not matter whether or not any complete successfully, the behavior is the same.
3. Delete the `Object`. Observe that the `Job` is deleted but all `Pods` for the job are still there
4. Create this `Object`

```yaml
kind: Object
metadata:
  name: test-deploy
spec:
  references:
  forProvider:
    manifest:
      apiVersion: apps/v1
      kind: Deployment
      metadata:
        name: nginx-deployment
        namespace: luna-infra
        labels:
          app: nginx
      spec:
        replicas: 3
        selector:
          matchLabels:
            app: nginx
        template:
          metadata:
            labels:
              app: nginx
          spec:
            containers:
            - name: nginx
              image: nginx:1.14.2
              ports:
              - containerPort: 80
```

2. Allow it to create its pods
3. Delete the `Object`. Observe that the `Deployment` is deleted and all `Pods` for the deployment are deleted

Crossplane version: 1.19.0
Provider version: 0.17.0
Kubernetes version: v1.30.9-gke.1046000
Kubernetes distribution: Google Kubernetes Engine (GKE)

alanrichman:

Fixes #342

When an `Object` resource is deleted, dependent resources (particularly Pods owned by Jobs) are not being properly cleaned up, resulting in orphaned Pods that remain in the cluster for a long time. In my experience it is about a week.

I have:

I followed the same procedure from the original issue with the current latest version of the provider (v1.2.1) and with the modified version from my fork. I validated that before the change Object deletion removed the Job but not its Pods, after the change Object deletion removed the Job and Pods.

### DIFF
--- a/examples/namespaced/provider/config.yaml
+++ b/examples/namespaced/provider/config.yaml
@@ -11,6 +11,7 @@ apiVersion: kubernetes.m.crossplane.io/v1alpha1
 kind: ClusterProviderConfig
 metadata:
   name: kubernetes-provider
+  namespace: default
 spec:
   credentials:
     source: Secret
--- a/examples/namespaced/provider/config.yaml
+++ b/examples/namespaced/provider/config.yaml
@@ -11,7 +11,6 @@ apiVersion: kubernetes.m.crossplane.io/v1alpha1
 kind: ClusterProviderConfig
 metadata:
   name: kubernetes-provider
-  namespace: default
 spec:
   credentials:
     source: Secret
--- a/README.md
+++ b/README.md
@@ -79,20 +79,20 @@ locally running controller.
       ```
       SA=$(kubectl -n crossplane-system get sa -o name | grep provider-kubernetes | sed -e 's|serviceaccount\/|crossplane-system:|g')
       kubectl create clusterrolebinding provider-kubernetes-admin-binding --clusterrole cluster-admin --serviceaccount="${SA}"
-      kubectl apply -f examples/provider/config-in-cluster.yaml
+      kubectl apply -f examples/namespaced/provider/config-in-cluster.yaml
       ```
   1. If provider kubernetes running outside the cluster (e.g. running locally with `make run`)
 
       ```
       KUBECONFIG=$(kind get kubeconfig --name local-dev | sed -e 's|server:\s*.*$|server: http://localhost:8081|g')
       kubectl -n crossplane-system create secret generic cluster-config --from-literal=kubeconfig="${KUBECONFIG}"
-      kubectl apply -f examples/provider/config.yaml
+      kubectl apply -f examples/namespaced/provider/config.yaml
       ```
 
-1. Now you can create `Object` resources with provider reference, see [sample object.yaml](examples/object/object.yaml).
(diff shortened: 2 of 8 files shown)


phisco:
This would mean everything gets deleted with foreground deletion, which would be a substantial behavior change. What if we made it configurable through the Object spec, defaulting to Background to preserve the current behavior?

alanrichman:
> This would mean everything gets deleted with foreground deletion, which would be a substantial behavior change. What if we made it configurable through the Object spec, defaulting to Background to preserve the current behavior?

This is a great idea! I am not sure why I did not get a notification from GH that you left this comment but I can take a stab at implementing this on my branch.

alanrichman:
@phisco Just pushed some new commits adding this as a configurable value in the Object's spec

alanrichman:
@phisco Just bumping this, GitHub's PR search bug has rendered me unable to find this one through regular means so I had to trace it back through the issues.

examples/namespaced/provider/config.yaml
```diff
@@ -11,6 +11,7 @@ apiVersion: kubernetes.m.crossplane.io/v1alpha1
 kind: ClusterProviderConfig
 metadata:
   name: kubernetes-provider
+  namespace: default
```
phisco: 🤔 ClusterProviderConfig is cluster scoped
```suggestion
```

phisco (approved):
just some questions around namespaced/cluster scoped, otherwise LGTM 🎉 sorry for the late reply, @alanrichman!

examples/namespaced/provider/config.yaml
alanrichman: Good catch, thank you

alanrichman:
> just some questions around namespaced/cluster scoped, otherwise LGTM 🎉 sorry for the late reply, @alanrichman!

No problem! Applied the change, and I appreciate the review

Issue #482, opened 2026-06-03
Remove managed fields from Object status


blut:
### What problem are you facing?
Having a high number of Objects on a kubernetes cluster uses much more Memory than required.

<img width="1306" height="320" alt="Image" src="https://github.com/user-attachments/assets/0b110690-bf7d-4d7a-b8cd-ef350307a390" />

This significantly affects Kubernetes API performance.

### How could Crossplane help solve your problem?

Allow to remove `status.atProvider.manifest.metadata.managedFields` for Objects.

Option 1: Automatically, always drop the managed fields from the status.
Option 2: Implement option / toggle to drop managed fields from the status.
Option 3: Implement parameter for Object to allow dropping managed fields from Object status.

jonasz-lasut:
Thank you @blut for opening the issue, this feature is available on main branch and will be a part of the next release (implemented in #483).

PR #483, opened 2026-06-05
feat: remove managed fields

blut:
Removes managed fields from object status

Fixes # https://github.com/crossplane-contrib/provider-kubernetes/issues/482

I have:

### DIFF
--- a/internal/controller/cluster/object/object.go
+++ b/internal/controller/cluster/object/object.go
@@ -152,7 +152,7 @@ type ResourceSyncer interface {
 }
 
 // Setup adds a controller that reconciles Object managed resources.
-func Setup(mgr ctrl.Manager, o controller.Options, sanitizeSecrets bool, pollJitterPercentage uint, legacyCSAFieldManagers []string) error { // nolint:gocyclo // Too many branches due to alpha features, hopefully we can clean them up after we graduate them.
+func Setup(mgr ctrl.Manager, o controller.Options, sanitizeSecrets bool, removeManagedFields bool, pollJitterPercentage uint, legacyCSAFieldManagers []string) error { // nolint:gocyclo // Too many branches due to alpha features, hopefully we can clean them up after we graduate them.
 	name := managed.ControllerName(v1alpha2.ObjectGroupKind)
 	l := o.Logger.WithValues("controller", name)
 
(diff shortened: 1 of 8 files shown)


blut:
I've built and deployed the provider with this version.
Behaviour is as expected and reducing etcd memory usage by ~30%.

I don't see what we did for e2e tests to fail.

internal/controller/cluster/object/object.go
```diff
@@ -518,6 +518,7 @@ func (c *external) setAtProvider(obj *v1alpha2.Object, observed *unstructured.Un
 
 	// sanitize/mutate only the copied object
 	sObserved := observed.DeepCopy()
+	sObserved.SetManagedFields(nil)
```
bobh66: Probably should put this behind a feature flag just to be safe

bobh66 (changes requested):
Thanks for the PR!  I think the e2e test failures were on main - if you rebase you should pick up the fix that just merged.
I agree with the changes but I think to be safe we should put them behind a feature flag in case someone is using the existing content in `status.atProvider` for something.  A CLI flag is probably fine - `Object` control is probably not needed at this point.   Thanks

internal/controller/cluster/object/object.go
blut: Thanks for your input. I've rebased and added the feature flag `strip-managed-fields` to enable this behaviour.

blut:
> Thanks for the PR! I think the e2e test failures were on main - if you rebase you should pick up the fix that just merged. I agree with the changes but I think to be safe we should put them behind a feature flag in case someone is using the existing content in `status.atProvider` for something. A CLI flag is probably fine - `Object` control is probably not needed at this point. Thanks

Hi, thanks for the input!

I've added the --strip-managed-fields flag and rebased on main!

bobh66:
@blut thanks for the updates.  My only concern is with "strip" vs "remove" - is "remove" easier to understand for someone for whom English isn't their first language?  Naming is hard - I can go either way.

blut:
> @blut thanks for the updates. My only concern is with "strip" vs "remove" - is "remove" easier to understand for someone for whom English isn't their first language? Naming is hard - I can go either way.

As English is not my first language, I disagree.
I've updated the flag and code to --remove-managed-fields

blut:
bump

bobh66 (approved):

PR #499, opened 2026-06-19
feat(nebius): add in-process service account identity auth for Nebius Cloud

jonasz-lasut:

Add a NebiusServiceAccountCredentials identity that authenticates to Nebius Managed Kubernetes (mk8s) clusters by minting IAM tokens in-process from a service account's credentials, rather than relying on an exec credential plugin.

- Add the nebius auth package wrapping a gosdk-backed token source
- Register and wire the identity into the kube client builder
- Cache token sources per credentials so the background-refreshing gosdk client is reused across reconciles instead of leaked
- Add ProviderConfig examples for the new identity type

I have:

Tested manually against a running Nebius managed kubernetes cluster


pkg/kube/client/nebius/store.go
```diff
@@ -0,0 +1,55 @@
...
+	if ok {
+		return src, nil
+	}
+
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	if src, ok := s.sources[key]; ok {
+		return src, nil
+	}
+	src, err := build(ctx)
```
erhancagirici, line 43: nit, potential overengineering, just for capturing here, we can follow-up later if we really need it:

Since this token minting happens while write-lock is held for the whole store, this also blocks other Nebius token operations , i.e. multiple independent Nebius-related reconciliations are serialized. 
Real-world effect probably not that big, as I assume credential rotations are sparse.
but potentially degrade reconciliation times of other Nebius creds when Nebius auth endpoint is slow and 

One idea might be optimistically minting a token before taking the write-lock (after cache-miss). 
and/or having another keyed lock for token retrieval (like https://pkg.go.dev/golang.org/x/sync/singleflight). This might have other tradeoffs as well, so lets proceed with current state for now.

pkg/kube/client/nebius/store.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		return nil, err
+	}
+	s.sources[key] = src
```
erhancagirici, line 47: nit, just noting, follow:
Sources are never invalidated and the cache itself always grows. Unless nebius go sdk has some other internal mechanism, sdk clients will pile up.
Though, in practice this might be negligible assuming the credential rotations are sparse.

Later, we can consider/introduce a cache invalidation mechanism (e.g. some TTL + utilize `nebiusSDK.Close()` on token sources)

pkg/kube/client/nebius/transport.go
```diff
@@ -0,0 +1,64 @@
...
+				req.Body.Close() //nolint:errcheck
+			}
+		}()
+	}
+
+	tkn, err := t.Source.BearerToken(req.Context())
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot get Nebius IAM token")
+	}
+	req2 := cloneRequest(req) // per RoundTripper contract
```
erhancagirici: nit: if I'm not missing anything, maybe utilizing `req.Clone(req.Context())`would be simpler to maintain

erhancagirici (approved):
@jonasz-lasut thanks for this addition 🎉 LGTM already. Left a few non-blocking notes.

pkg/kube/client/nebius/transport.go
jonasz-lasut: let me test it

pkg/kube/client/nebius/store.go
jonasz-lasut: I was thinking about cache invalidation as well but like you've mentioned credential rotations are sparse, especially if you take into consideration that Nebius uses service account based auth and does not use static tokens, which would have to be regularly rotated.
I'd keep it like this for now and come back if we see community reporting issues with the resource consumption

pkg/kube/client/nebius/transport.go
jonasz-lasut: works the same, thank you for pointing it!

PR #502, opened 2026-06-24
Observer Only without diff and extra SSA call

ihakimis:

Fix Observe mode SSA validation failure for sparse manifests in both namespaced and cluster scopes.

When `managementPolicies: [Observe]` is set and SSA is enabled (`--enable-beta-server-side-apply`), the controller was unconditionally calling `GetDesiredState` during reconcile even for observe-only resources. For SSA, `GetDesiredState` performs a **dry-run server-side apply** against the Kubernetes API. This dry-run fails for manifests that intentionally omit `spec` — for example an `Endpoints` resource defined with only `apiVersion`, `kind`, and `metadata`.

The fix follows the existing structure of `handleObservation`, which already short-circuits with `isUpToDate = true` for observe-only resources. The desired state computation is moved inside `handleObservation` and is now only performed when the management policy actually requires managing the resource (i.e. includes `Create`, `Update`, or `All`).

**Changes:**
- `Observe()`: removed the unconditional `GetDesiredState` call; passes the raw `manifest` to `handleObservation` instead of a pre-computed desired state
- `handleObservation()`: computes desired state lazily — only when not observe-only. For observe-only resources, the function returns `isUpToDate = true` immediately, skipping any API interaction
- Tests (`object_test.go`): the `"Observe Only - up to date by default"` test case no longer sets `GetDesiredStateFn`. Because a nil function panics if called, this acts as a live assertion that `GetDesiredState` is never invoked for observe-only resources
- Both **namespaced** and **cluster** scopes are fixed identically

Fixes #431 

I have:

- Existing unit tests updated to assert `GetDesiredState` is never called for observe-only resources (nil `GetDesiredStateFn` panics on invocation — test passing proves it is not called)
- Manually reproduced with an `Endpoints` object CR using only `apiVersion`, `kind`, and `metadata` in `forProvider.manifest` with SSA enabled and `managementPolicies: [Observe]` — confirms the controller no longer errors during reconcile
- All existing `TestObserve` cases continue to pass for both namespaced and cluster scopes, confirming non-observe-only flows are unaffected

### DIFF
--- a/examples/cluster/object/object-csa-migration.yaml
+++ b/examples/cluster/object/object-csa-migration.yaml
@@ -14,7 +14,7 @@ metadata:
     # unpauses the MR to ensure it gets reconciled first with CSA.
     crossplane.io/paused: "true"
     uptest.upbound.io/post-assert-hook: testhooks/validate-ssa-migration.sh
-    uptest.upbound.io/timeout: "60"
+    uptest.upbound.io/timeout: "180"
 spec:
   forProvider:
     manifest:
--- a/examples/cluster/object/testhooks/validate-ssa-migration.sh
+++ b/examples/cluster/object/testhooks/validate-ssa-migration.sh
@@ -65,6 +65,30 @@ current_provider_args=$(${KUBECTL} get pods -A -l pkg.crossplane.io/provider=pro
 echo " ⛭ current provider args: $current_provider_args"
 echo " ☑️ Enabled server-side apply in provider..."
 
+# Wait for the provider to reconcile the object after SSA is enabled.
+# The provider needs time to observe the object, upgrade field managers, and
+# apply the desired state which removes the unwanted label and last-applied annotation.
+SSA_RECONCILE_TIMEOUT=60
+SSA_RECONCILE_START=$(date +%s)
+echo " ⏳ Waiting up to ${SSA_RECONCILE_TIMEOUT}s for SSA migration to complete..."
+while true; do
+  NOW=$(date +%s)
+  ELAPSED=$((NOW - SSA_RECONCILE_START))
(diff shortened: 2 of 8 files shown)


bobh66:
@ihakimis thanks for the PR!  Can you take care of the DCO and the e2e-test failure?

ihakimis:
@bobh66 PR is ready! all tests passed!

ihakimis:
Hi Team @bobh66, 
Any update when this fix will be merged?

jcogilvie:
As it stands, this is going to be in conflict with #524 which will implement the `UpToDate` condition that adds a field-for-field diff if the object we're observing is different from the one we've supplied.  However, i think I see a way to reconcile both approaches, which is to have the diff gracefully degrade if it hits the schema validation error that you're trying to avoid for sparse manifests.  If I can make that work, then #524 will supersede this change, but fix the issue at hand.

I'm actively working with @bobh66 on it.  

Sorry to swoop in here with this at the last minute!  It shouldn't be long before mine's ready.

bobh66:
Is the need for the SSA call caused by the possibility that there could be write-only fields in the Object?  I'm wondering if this case is specific to provider-kubernetes or if there is a more global need to distinguish between references to an existing resource that we can't or don't want to specify any state for (`spec.forProvider.manifest.spec` could be empty in this scenario, or `spec.forProvider` in most other providers) vs the case where we want to compare what we think the resource should look like against what it really looks like.

I was thinking that in the "reference" Observe-only case we could provide only the external-name and leave `spec.forProvider` empty, to indicate that the provider should read the external resource (if it exists) but not do any comparisons.  If `spec.forProvider` is populated then it would be compared against the actual resource and a delta made available.

PR #480, opened 2026-05-29
feat: update Crossplane runtime to v2.3.1

patrickleet:
### Description of your changes

Updates provider-kubernetes to crossplane-runtime/v2 v2.3.1 and Go 1.26.3.

This also updates Crossplane core API imports to `github.com/crossplane/crossplane/apis/v2/core/v2`. Generated code still aliases this package as `xpv1`, but the imported package is the v2 API package.

The dependency bump required refreshing generated code and CRDs, plus small compatibility fixes for the updated Kubernetes/controller-runtime dependency graph.

Context: this is also needed by the companion provider-helm runtime upgrade. provider-helm depends on `github.com/crossplane-contrib/provider-kubernetes`, I ran local validation/tests of the provider-helm crossplane-runtime/v2 v2.3.1 with this provider-kubernetes branch as a test.

I have:

- [x] Read and followed the [contribution process].
- [x] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

Validated locally with:

- `make reviewable test`
- `make vendor.check`
- `git diff --check`

[contribution process]: https://git.io/fj2m9

PR #506, opened 2026-07-02
chore: use pflag ParseErrorsAllowlist and bump pflag to v1.0.10

jcogilvie:

pflag renamed `ParseErrorsWhitelist` to `ParseErrorsAllowlist` in v1.0.7 and marked the old name deprecated. Bump the direct `github.com/spf13/pflag` dependency to v1.0.10 and switch the Azure kubelogin flag set in `pkg/kube/client/azure/azure.go` to the non-deprecated name. `ParseErrorsWhitelist` remains an alias in v1.0.10, so this is purely a cleanup / deprecation avoidance.

This is extracted from #480 as a small precursor to the crossplane-runtime v2.3.1 upgrade so the runtime PR stays focused on the runtime change itself. Original change authored by @patrickleet; that authorship is preserved in the commit.

Fixes #

I have:

- `make reviewable test` (locally, on Go 1.26.4)
- Verified against `github.com/spf13/pflag@v1.0.10` source that `ParseErrorsWhitelist` is preserved as a type alias / deprecated field, so no downstream breakage.


bobh66 (approved):
LGTM - thanks!

PR #507, opened 2026-07-02
fix(ssa): strip API root prefix from OpenAPI v3 ServerRelativeURL

jcogilvie:

When the Kubernetes API server is exposed behind a URL path prefix (e.g. a proxy or an API server hosted under `/k8s/api/...`), the paths returned by the `/openapi/v3` discovery endpoint include that prefix. `pkg/kube/client/ssa/cache/extractor/caching_unstructured_extractor.go` then failed to detect them as OpenAPI v3 URLs because it compared `ServerRelativeURL` against a fixed `/openapi/v3` prefix, and the URL passed to the OpenAPI client no longer matched the client's expected shape.

Fix: compute the API root prefix from the REST client itself (via `rc.Get().AbsPath("/").URL().Path`) and strip it from `ServerRelativeURL` before both the `/openapi/v3` detection and downstream use.

This is extracted from #480 as a small precursor to the crossplane-runtime v2.3.1 upgrade so the runtime PR stays focused on the runtime change itself. Original change authored by @patrickleet; that authorship is preserved in the commit.

Fixes #

I have:

- `make reviewable test` (locally, on Go 1.26.4)
- Verified the fix is a no-op for clusters served at the root: `rc.Get().AbsPath("/").URL().Path` is `/`, so `rootPrefix` becomes an empty string after `TrimSuffix`, and the subsequent `TrimPrefix` is a no-op.

### DIFF
--- a/pkg/kube/client/ssa/cache/extractor/caching_unstructured_extractor.go
+++ b/pkg/kube/client/ssa/cache/extractor/caching_unstructured_extractor.go
@@ -70,6 +70,7 @@ func discoveryPaths(ctx context.Context, rc rest.Interface) (map[string]OpenAPIG
 		return nil, nil, err
 	}
 
+	rootPrefix := strings.TrimSuffix(rc.Get().AbsPath("/").URL().Path, "/")
 	oapiPathsToGV := map[string]OpenAPIGroupVersion{}
 	oapiPathsToETags := map[string]string{}
 	for path, oapiGV := range discoMap.Paths {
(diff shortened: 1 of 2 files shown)


pkg/kube/client/ssa/cache/extractor/caching_unstructured_extractor.go
```diff
@@ -70,13 +70,15 @@ func discoveryPaths(ctx context.Context, rc rest.Interface) (map[string]OpenAPIG
...
 
+	rootPrefix := strings.TrimSuffix(rc.Get().AbsPath("/").URL().Path, "/")
 	oapiPathsToGV := map[string]OpenAPIGroupVersion{}
 	oapiPathsToETags := map[string]string{}
 	for path, oapiGV := range discoMap.Paths {
 		parse, err := url.Parse(oapiGV.ServerRelativeURL)
 		if err != nil {
 			return nil, nil, err
 		}
+		oapiGV.ServerRelativeURL = strings.TrimPrefix(oapiGV.ServerRelativeURL, rootPrefix)
```
bobh66, line 81: Should there be a test for this to cover both the default case and the scenario that was causing the error?

bobh66 (approved):
LGTM - thanks!

PR #526, opened 2026-08-03
Log a diff when Object drift is detected at debug level

### LINKED ISSUE #422: CPU Usage Spike
We’re seeing problematic behavior with the provider-kubernetes when a composite resource is in an unhealthy state. In this scenario, it seems to trigger a continuous reconciliation loop for the Object—around ~5 reconcile requests per second. This leads to significant CPU spikes and eventually causes the pod to hit its resource limits.We’re currently running v0.16.0. Is there a way to tune or change this reconciliation behavior?

pujitha24:

When the Object controller decides a managed resource is not up-to-date, it
previously logged nothing at Debug level about *why* — only that it wasn't
up-to-date. A user on this issue reported that persistent drift (in their
case caused by an empty-string label in the manifest) was driving frequent
reconciliation and high CPU usage, and that they had no way to see the root
cause from the provider's debug logs, forcing trial-and-error to diagnose it.

This PR adds a Debug-level log line in `handleObservation` (both the cluster
and namespaced Object controllers) that includes a structural diff (via
`github.com/google/go-cmp`, already a direct dependency) between the observed
and desired state whenever the resource is found not up-to-date. This is a
diagnostics-only change — it does not alter `ExternalObservation` values,
reconcile cadence, or backoff behavior, and does not by itself address the
reconciliation-frequency/CPU-usage question raised in the issue (that would
need separate design discussion). It only makes it possible to root-cause
persistent drift from existing debug logs.

Fixes #

I have:

- [ ] Read and followed Crossplane's [contribution process].

Ran `go build ./...` (clean) and
`go test ./internal/controller/cluster/object/... ./internal/controller/namespaced/object/...`
(all tests pass), including a new `NotUpToDateNoObservedState` test case
added to both packages' `TestObserve` table that drives `handleObservation`
with a nil observed state to exercise the new log call and confirm it does
not panic. Also ran `go vet` and `gofmt -l` on the changed files (clean),
and `make lint` (golangci-lint), which reports no new findings in either
changed file — the pre-existing failures it reports are in unrelated files
and predate this change.

Fixes #422

### DIFF
--- a/internal/controller/cluster/object/object.go
+++ b/internal/controller/cluster/object/object.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/google/cel-go/cel"
 	celtypes "github.com/google/cel-go/common/types"
+	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
@@ -777,6 +778,8 @@ func (c *external) handleObservation(ctx context.Context, obj *v1alpha2.Object, 
 			ConnectionDetails: cd,
 		}, nil
 	}
+
+	c.logger.Debug("Not up to date", "diff", cmp.Diff(last, desired))
 
 	return managed.ExternalObservation{
 		ResourceExists:   true,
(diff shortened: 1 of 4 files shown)


pujitha24:
Closing this to keep my open PR queue manageable - it has been open a while without review, and I would rather not leave stale PRs sitting in your queue. The change itself still applies; happy to reopen and rebase if it is useful to you.

PR #528, opened 2026-08-04
Support Google service account impersonation and delegation chains for GKE authentication

### LINKED ISSUE #527: Support Google service account impersonation and delegation chains for GKE authentication

When provider-kubernetes authenticates to GKE using Google Application Credentials, it can only use the configured base identity directly.

Some environments require the provider's Workload Identity or supplied credentials to impersonate a dedicated Google service account before accessing the cluster. Organizations may also require one or more intermediate service accounts due to IAM boundaries, using an impersonation delegation chain.

Without this support, users must grant the provider's base identity direct access to every target cluster, which weakens isolation and complicates centralized IAM management.

Add optional Google service account impersonation configuration to `ProviderConfig` identities:

```yaml
identity:
  type: GoogleApplicationCredentials
  source: InjectedIdentity
  impersonateServiceAccount:
    name: target-sa@example-project.iam.gserviceaccount.com
    delegates:
      - intermediate-sa@example-project.iam.gserviceaccount.com
```

Expected behavior:

* Use the configured or injected Google identity as the base credential.
* Impersonate the target service account through Google IAM Credentials.
* Pass delegates in the configured order.
* Use the resulting impersonated token to authenticate to GKE.
* Support both cluster-scoped and namespaced ProviderConfigs.
* Fail closed when impersonation is explicitly configured but invalid.
* Preserve existing behavior when impersonation is omitted.
* Document required `roles/iam.serviceAccountTokenCreator` grants.
* Include tests covering direct impersonation, delegation chains, and base-credential forwarding.

kostyaplis:

Adds support for Google service account impersonation — including delegation chains — to GoogleApplicationCredentials identities, so the provider can authenticate to GKE clusters as a narrowly-scoped target service account instead of its base identity.

Fixes #527  

**API** (`pkg/kube/config`): a new optional `identity.impersonateServiceAccount` block with a required name (the target service account email) and an optional ordered `delegates` chain. Added to all three ProviderConfig CRDs (cluster-scoped `kubernetes.crossplane.io`, and both `kubernetes.m.crossplane.io` variants).

**Behavior** (`pkg/kube/client/gke`): `WrapRESTConfig` now accepts an optional `*Impersonation`. The base token source — built from the injected identity or supplied credentials — signs the impersonation exchange via `impersonate.CredentialsTokenSource` with `option.WithTokenSource`, and the impersonated token is what authenticates to the cluster. The base source is passed explicitly so impersonation is never silently signed by Application Default Credentials.

Examples demonstrating the new field are included under `examples/cluster` and `examples/namespaced`.

I have:

- Unit tests for `gke.WrapRESTConfig` (`pkg/kube/client/gke/gke_test.go`) covering: access-token, JSON, and injected-identity credential paths; impersonation skipped when not configured; fail-closed error on an empty target; the delegation chain forwarded in order; and a regression guard asserting the impersonation exchange receives the token source built from the supplied credentials (not ADC) and that the transport signs with the impersonated token.
- Unit tests for the ProviderConfig-to-wrapper wiring (`pkg/kube/client/client_test.go`) covering the mapping of `name`/`delegates` into `gke.Impersonation` for both Secret-sourced and injected-identity credentials, and the nil case.
- Built, deployed, and tested against our GKE cluster fleet - verifying impersonation for both existing and newly created infrastructure.


awprice:
I have a related AWS proposal in #563. It needs an `assumeRoleChain` for `AWSWebIdentityCredentials`, so the two changes have the same API design question.

This PR adds `impersonateServiceAccount` directly under `identity`. In #563, I am proposing a nested block because `Identity` could otherwise end up with several fields that each apply to only one identity type.

Would it make sense to group these settings by cloud?

```yaml
identity:
  type: GoogleApplicationCredentials
  source: InjectedIdentity
  google:
    impersonateServiceAccount:
      name: target-sa@example-project.iam.gserviceaccount.com
      delegates:
        - intermediate-sa@example-project.iam.gserviceaccount.com
```

The AWS version would be:

```yaml
identity:
  type: AWSWebIdentityCredentials
  source: InjectedIdentity
  aws:
    assumeRoleChain:
      - roleARN: arn:aws:iam::123456789012:role/target-eks-access
```

Another option is to name the blocks after the identity types, such as `googleApplicationCredentials` and `awsWebIdentityCredentials`.

I would like this PR and #563 to use the same pattern before either change is merged. Would either nested form work here?

kostyaplis:
@awprice, thanks for you suggestion!
I’m happy to continue working on this PR. Hopefully, the maintainers can comment or suggest what shape the `identity` block should take.

Issue #420, opened 2026-01-06
Observe fails on Secret with stringData with server-side-apply


bobh66:
### What happened?
We enabled Server Side Apply on provider-kubernetes v0.18.0 and when we create a `Secret` using
`stringData` the observe step fails.

### How can we reproduce it?

This manifest:
```yaml
apiVersion: kubernetes.crossplane.io/v1alpha2
kind: Object
metadata:
  name: teststringdata
spec:
  deletionPolicy: Delete
  forProvider:
    manifest:
      apiVersion: v1
      kind: Secret
      metadata:
        name: teststringdata
        namespace: default
      stringData:
        password: DBPassword1?
        username: cmuser
  managementPolicies:
  - '*'
  providerConfigRef:
    name: local
  readiness:
    policy: SuccessfulCreate
  watch: false
```

results in this condition:

```yaml
apiVersion: kubernetes.crossplane.io/v1alpha2
kind: Object
metadata:
  annotations:
    crossplane.io/external-create-pending: "2026-01-06T18:13:04Z"
    crossplane.io/external-create-succeeded: "2026-01-06T18:13:04Z"
    crossplane.io/external-name: teststringdata
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"kubernetes.crossplane.io/v1alpha2","kind":"Object","metadata":{"annotations":{},"name":"teststringdata"},"spec":{"deletionPolicy":"Delete","forProvider":{"manifest":{"apiVersion":"v1","kind":"Secret","metadata":{"name":"teststringdata","namespace":"default"},"stringData":{"password":"DBPassword1?","username":"cmuser"}}},"managementPolicies":["*"],"providerConfigRef":{"name":"local"},"readiness":{"policy":"SuccessfulCreate"},"watch":false}}
  creationTimestamp: "2026-01-06T18:13:04Z"
  finalizers:
  - finalizer.managedresource.crossplane.io
  generation: 1
  name: teststringdata
  resourceVersion: "737783"
  uid: 2b16c36b-3eef-44e1-b7b8-58def8458083
spec:
  deletionPolicy: Delete
  forProvider:
    manifest:
      apiVersion: v1
      kind: Secret
      metadata:
        name: teststringdata
        namespace: default
      stringData:
        password: DBPassword1?
        username: cmuser
  managementPolicies:
  - '*'
  providerConfigRef:
    name: local
  readiness:
    policy: SuccessfulCreate
  watch: false
[...]

Issue #418, opened 2025-12-17
Cannot get observed state at secret with empty data


CsDenes:
### What happened?
After creating a Kubernetes Secret with an empty data field, the observer fails with the following warning, and the Secret cannot be managed afterward. This issue appeared after enabling the `-enable-server-side-apply`.

```
cannot get observed state: failed calling ExtractInto for unstructured: unable to convert managed fields for provider-kubernetes/test-secret to unstructured, expected map, got <nil>
```

### How can we reproduce it?
Create an object with the following manifest, where `data` field is specified but empty.

```yaml
apiVersion: kubernetes.crossplane.io/v1alpha2
kind: Object
metadata:
  name: test-secret
spec:
  deletionPolicy: Delete
  forProvider:
    manifest:
      apiVersion: v1
      data:
      kind: Secret
      metadata:
        name: test-secret
        namespace: default
  managementPolicies:
  - '*'
  providerConfigRef:
    name: default
```

### What environment did it happen in?
Crossplane version: v1.19.0
provider-kubernetes version: v0.18.0  (with --enable-server-side-apply)
Kubernetes version: v1.34.0

jonasz-lasut:
Thank you @CsDenes for opening the issue — this is the same root cause as #420: the SSA extractor fails with `ExtractInto: expected map, got <nil>` whenever the provider's field manager only owns fields absent from the observed object (empty `data` here, `stringData` there); see the analysis in #420 and upstream kubernetes/kubernetes#141153. Closing as a duplicate to consolidate tracking in #420.

PR #534, opened 2026-08-05
docs: document write-only field limitation in Object manifests

jonasz-lasut:

Adds a "Known limitations" README section documenting that write-only fields (canonically `Secret.stringData`) cannot be correctly reconciled in `Object` manifests, per the analysis in #420:

- under server-side apply, a write-only field appears in neither the observed extraction nor the dry-run desired state, so post-creation changes to it produce no diff and are silently not applied;
- when the provider's field manager owns only such fields, the managed-fields extraction fails with `ExtractInto ... expected map, got <nil>` and the `Object` is stuck in `ReconcileError` (#420, #418).

Recommends round-trippable fields instead (base64 `data` for Secrets). This documents the limitation as suggested in the #420 thread; the issues themselves stay open for the underlying extractor fix.

I have:

Documentation-only change; rendered locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,37 @@ spec:
   package: xpkg.crossplane.io/crossplane-contrib/provider-kubernetes:v1.0.0
 ```
 
+## Known limitations
+
+### Write-only fields cannot be reconciled
+
+Some Kubernetes fields are "write-only": the API server accepts them but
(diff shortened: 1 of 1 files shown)


sergenyalcin (approved):
Thanks @jonasz-lasut LGTM!

PR #543, opened 2026-08-19
Fix per reconcile discovery

### LINKED ISSUE #541: Memory/CPU regression under watch load: full aggregated discovery runs on every reconcile
**Summary:** On a cluster with a large CRD population, `provider-kubernetes` `main`
uses roughly **2.8x the steady memory and ~5x the CPU** of `v1.2.1` under the same
`watch: true` load. It looks like a leak but is not one: it is a **transient ~2 GiB
heap spike per reconcile round** that Go returns to the OS lazily (`MADV_FREE`), so
the working set decays slowly and *looks* retained. The provider builds a **fresh
controller-runtime client (and therefore a fresh `RESTMapper`) on every `Connect`**,
and controller-runtime `>= v0.20.0` primes that mapper with **full aggregated
discovery** - one `GET /api` + `GET /apis` returning *every* CRD in the cluster - on
each build. `v1.2.1` ships controller-runtime `v0.19.0`, which fetched discovery
**lazily, per requested group/version**. The behavior flip is upstream
controller-runtime [#2901][cr2901] (issue [#2897][cr2897]), first released in
`controller-runtime v0.20.0`.

## Environment

- Provider: `crossplane-contrib/provider-kubernetes` `main` vs `v1.2.1` (release).
- controller-runtime: **`v0.23.1`** on `main` vs **`v0.19.0`** on `v1.2.1` - this
  bump is the pivot. The relevant behavior change landed in `v0.20.0`.
- Crossplane: UXP with the external package runtime; `kind` cluster.
- Kubernetes: aggregated discovery is beta/default since `1.27`, so any reasonably
  modern apiserver takes the aggregated path.
- Cluster shape that makes it visible: **~1000 CRDs**, 300 `Object`s with `watch: true`, a `ClusterProviderConfig` with credential
  `source: Secret` (kubeconfig) so every reconcile reads credentials. **Invisible on
  a bare cluster** - a large CRD population is required to trigger it.

## What happens

1. Each `Object` reconcile calls `Connect`, which builds a new controller-runtime
   client for the target cluster via `client.New(rc, client.Options{})` in
   [`pkg/kube/client/client.go:93`][clientnew] (`IdentityAwareBuilder.KubeForProviderConfig`,
   fresh `rest.Config` -> fresh `RESTMapper`).
2. controller-runtime `>= v0.20.0` initializes that `RESTMapper` through the
   `AggregatedDiscoveryInterface`: on the first mapping lookup it issues `GET /api` +
   `GET /apis`, which on an aggregated-discovery apiserver returns **all 1572 CRDs in
   one multi-MB payload** that is decoded and cached wholesale. `v0.19.0` did not -
   it lazily fetched only the specific `group/version` the reconcile asked for
   (`ServerGroups()` + `ServerResourcesForGroupVersion(<group>/<version>)`).
3. So every reconcile round of 300 `Object`s on `main` performs a full,
   multi-megabyte discovery decode 300 times over. The working set balloons to
   ~2 GiB during the round, then decays slowly afterward.

## Root cause (confirmed - code + apiserver metrics + heap profiling)

The mechanism is **per-reconcile full aggregated discovery**, not a retained
allocation. The provider builds a fresh client per `Connect` in both versions - that
is unchanged. What changed is what *building a client costs*: controller-runtime
`v0.19` fetched discovery lazily per group/version; controller-runtime `>= v0.20`
always primes via aggregated discovery.

- **Confirmed (code):** the call site is a per-`Connect` `client.New` with no cached
  mapper ([`client.go:93`][clientnew]). Upstream, controller-runtime `v0.19.7`
  restmapper uses `ServerGroups()` + per-group `ServerResourcesForGroupVersion`;
  `>= v0.20.0` uses `GroupsAndMaybeResources()` (the aggregated `/api` + `/apis`
  path) and caches all groups/resources at once.
- **Confirmed (apiserver metrics):** an `apiserver_request_total` diff over one
  300-reconcile round shows **614 + 614 aggregated discovery `GET`s on `main`** vs
  **3 + 3 on `v1.2.1`**. `APPLY` counts are identical across the two builds - the
  delta is purely discovery, not extra writes.
- **Confirmed (live heap):** during a round, `main` shows `heap_sys`
  `2031-2276 MiB` / `heap_inuse` peak `1964 MiB` vs `v1.2.1` `538-586` / `445`. An
  idle `main` pod sits at `heap_sys 2.15 GiB` with only `285 MiB` live and
  **`1.54 GiB` madvised** (freed to the OS, not yet reclaimed).

## Proposed fix - cache the client per credential identity

**Confirmed (prototype, measured):** an **LRU client cache in `IdentityAwareBuilder`**,
keyed by the **SHA-256 of the extracted credential material** and guarded by
**singleflight** so concurrent reconciles for the same identity build the client
once. Discovery is then primed once per credential, not once per reconcile.

Credential rotation stays safe: all identity wrappers install self-refreshing token
sources behind `rc.Wrap` transports, so the bearer token refreshes underneath the
cached client. The cache key is the *extracted credential bytes*, so a genuinely new
credential yields a new key and a new client, while a rotating token on the same
credential keeps reusing the live client - no stale-auth window. The same fix also
closes the latent `tlsTransportCache` leak for varying-target credentials.

## Related upstream issues & fix

- **[kubernetes-sigs/controller-runtime#2897][cr2897]** - *"Restmapper: Use
  aggregated discovery if available"*. Maintainer framing: *"We've built a lazy
  restmapper that only fetches mappings when needed... upstream built aggregated
  discovery which allows to fetch all restmappings with a single api call."*
- **[kubernetes-sigs/controller-runtime#2901][cr2901]** - the implementing PR
  (merged 2024-09-23, first released in **`v0.20.0`**). From the description:
  *"during startup, we will always do a request to `/api` and `/apis`... the only
  case where this is worse is if the client interacts with only one GroupVersion."*
  Upstream optimized for **request count on a long-lived mapper**; it did not
  anticipate a mapper rebuilt *per reconcile*, which is what turns "one extra
  request" into a multi-MB decode of every CRD, 300x per round.
- **[kubernetes-sigs/controller-runtime#3441][cr3441]** - the `tlsTransportCache`
  transport leak, closed *not-planned* (inherited client-go behavior).

## Reference

- [`pkg/kube/client/client.go:88-93`][clientnew] - the per-`Connect` `client.New`
  call site.
- [crossplane-contrib/provider-kubernetes#425][pr425] - metadata-only watch
  informers, merged; already in `main`, so it is **not** the regression.

[cr2897]: https://github.com/kubernetes-sigs/controller-runtime/issues/2897
[cr2901]: https://github.com/kubernetes-sigs/controller-runtime/pull/2901
[cr3441]: https://github.com/kubernetes-sigs/controller-runtime/issues/3441
[pr425]: https://github.com/crossplane-contrib/provider-kubernetes/pull/425
[clientnew]: https://github.com/crossplane-contrib/provider-kubernetes/blob/9674f8ece416994fb1626e2f4ce706b6cbe60062/pkg/kube/client/client.go#L88-L93

jonasz-lasut:

Fixes #541

**Summary:** every reconcile built a fresh controller-runtime client for the target cluster, and since controller-runtime `v0.20.0` a fresh client's `RESTMapper` primes itself with *full aggregated discovery* (`GET /api` + `GET /apis`, every CRD in the cluster) on its first lookup. On CRD-heavy clusters that is a multi-megabyte decode per reconcile, which is the ~3x memory / ~5x CPU regression under `watch` load described in #541 (root cause, upstream references and the original measurements live there). This PR caches the built client per credential set so discovery is primed once per credential set instead of once per reconcile.

Two commits, reviewable separately:

1. `fix(aws): issue EKS tokens with the request context` - `eksTokenSource` stored the context of the reconcile that built the REST config and presigned every later `GetCallerIdentity` with it ([`aws.go:125`](https://github.com/jonasz-lasut/provider-kubernetes/blob/c56b5e3/pkg/kube/client/aws/aws.go#L125), [`:161`](https://github.com/jonasz-lasut/provider-kubernetes/blob/c56b5e3/pkg/kube/client/aws/aws.go#L161)). crossplane-runtime cancels that context when the reconcile returns, so anything that outlives the reconcile (today the `watch` informers, with this PR the cached client) refreshed IRSA credentials against a dead context. The token source now takes `req.Context()`. Standalone fix, safe on its own.
2. `fix: cache per-ProviderConfig clients to avoid discovery per reconcile` - the cache itself.

#### What the cache does

- `IdentityAwareBuilder` keeps an LRU (`k8s.io/utils/lru`, already a dependency) of built clients keyed by a SHA-256 digest of everything that shapes the `rest.Config`: credential source and kubeconfig bytes, identity type/source and identity credential bytes, and the AWS cluster name ([`client.go:150-183`](https://github.com/jonasz-lasut/provider-kubernetes/blob/c56b5e3/pkg/kube/client/client.go#L150-L183), fields length-prefixed at [`:187`](https://github.com/jonasz-lasut/provider-kubernetes/blob/c56b5e3/pkg/kube/client/client.go#L187)). A changed credential is a new key and a new client; a rotating token behind the same credential keeps the live client, because every identity wrapper installs a self-refreshing token source behind `rc.Wrap`. `singleflight` collapses concurrent cold-cache builds of the same client.
- One builder is shared by the four controllers (cluster/namespaced x Object/ObservedObjectCollection) through `internal/controller.Options.ClientBuilder`, so the cache is shared too. Its bound is `--client-cache-size` (default `8`: one entry per distinct credential set, and most control planes manage the in-cluster API plus one or two clusters; `0` removes the bound, values above `math.MaxInt` are rejected at startup, and the builder treats a negative size as `0`). `Setup` and `SetupGated` fail without a builder instead of panicking on the first reconcile. Build and eviction are logged at debug level, never the key.
- The identity wrappers are built with `context.WithoutCancel(ctx)` ([`client.go:242`](https://github.com/jonasz-lasut/provider-kubernetes/blob/c56b5e3/pkg/kube/client/client.go#L242)) while Secret reads stay bound to the reconcile context. Without this, the Google `external_account`/`authorized_user` token sources (which store the context they were created with and do HTTP with it) would fail with `context canceled` permanently after their first access-token expiry once the client is cached; `TestKubeForProviderConfigOutlivesContext` reproduces exactly that against an httptest STS and fails if the detachment is removed.
- `rc.QPS = -1` for target clusters ([`client.go:236`](https://github.com/jonasz-lasut/provider-kubernetes/blob/c56b5e3/pkg/kube/client/client.go#L236)). **Behavior change, please look at this one:** a cached client is shared by every concurrent reconcile of its credential set, so client-go's per-client token bucket (`5 QPS / burst 10`, the default whenever a config sets none, which is the case for both the in-cluster and the kubeconfig path here) would have serialized them, where it previously applied per reconcile and never bit. Disabling client-side throttling keeps the effective pre-PR behavior for reconciles, but it applies to everything built from the returned config: the reconcile client, the watch informers and the server-side apply discovery client are all unthrottled on the client side now. `--max-reconcile-rate` and the target cluster's API Priority and Fairness bound the load instead; the README documents this. The returned `*rest.Config` is a copy so callers cannot mutate the cached one.
- `token.ReuseSourceStore` (Upbound) and `nebius.SDKStore` stay: they cache a different layer (the token source / SDK, keyed by identity credential only) and are still needed because the wrappers run before the cache lookup and two clusters sharing an identity share one token source.

**Confirmed (code, controller-runtime `v0.23.1`):** `apiutil` mapper holds a `discovery.AggregatedDiscoveryInterface` and primes all groups on the first lookup; `client.New`, `cache.New` and `discovery.NewDiscoveryClientForConfig` each `rest.CopyConfig` first and all per-call options (field owner, dry-run, force) are per request, so sharing one client across reconcilers is safe.

#### Benchmark

**Tested (kind `v1.36.1`, single node, Docker via colima on Apple Silicon macOS; Crossplane `v2.3.4`; provider built and deployed with `make local-deploy`, `--debug`, watches off, default `--max-reconcile-rate=100`; `main` vs this branch):** 1000 ballast CRDs (100 groups x 10 kinds), 300 `Object`s each managing a `ConfigMap` through `ClusterProviderConfig/kubernetes-provider` (`InjectedIdentity`), settle + 3 update rounds, provider sampled every 2 s with a 30 s cooldown per phase. Discovery requests are `apiserver_request_total{subresource="api"|"apis"}` from the API server; heap/RSS from the provider's `/metrics`; working set from the kubelet stats summary.

| phase | build | duration | discovery req | apiserver req | provider req | peak heap in use MiB | peak heap sys MiB | peak RSS MiB | peak working set MiB |
|---|---|---:|---:|---:|---:|---:|---:|---:|---:|
| settle | main | 6.26 s | 1278 | 7560 | 5619 | 834 | 954 | 1003 | 966 |
| round-1 | main | 2.31 s | 604 | 3180 | 2700 | 665 | 954 | 844 | 806 |
| round-2 | main | 2.25 s | 604 | 3181 | 2700 | 694 | 953 | 979 | 942 |
| round-3 | main | 2.48 s | 604 | 3182 | 2700 | 695 | 985 | 979 | 942 |
| settle | cache | 4.31 s | 6 | 5622 | 4392 | 106 | 124 | 161 | 102 |
| round-1 | cache | 2.41 s | 4 | 3136 | 2101 | 94 | 122 | 159 | 114 |
| round-2 | cache | 2.32 s | 4 | 3008 | 2101 | 92 | 133 | 171 | 131 |
| round-3 | cache | 2.45 s | 4 | 2584 | 2101 | 94 | 134 | 171 | 131 |

Steady state, average of rounds 1-3:

| metric | main | cache | change |
|---|---:|---:|---:|
| duration | 2.35 s | 2.39 s | ~same |
| discovery requests / round | 604 | 4 | -99.3% |
| apiserver requests / round | 3181 | 2909 | -9% |
| provider requests / round | 2700 | 2101 | -22% |
| peak heap in use | 685 MiB | 93 MiB | -86% |
| peak heap sys | 964 MiB | 130 MiB | -87% |
| peak RSS | 934 MiB | 167 MiB | -82% |
| peak working set | 897 MiB | 125 MiB | -86% |

`main` issues two discovery requests per `Object` per round (one per client build); the branch issues a constant 4. Round latency is unchanged at this scale: the regression is memory and CPU, not throughput.

#### Not in this PR

- No cache hit/miss metrics
- provider-helm imports this builder and inherits the fix on its next dependency bump; its own Helm SDK discovery path is separate.

I have:

- Unit tests (`make reviewable test` green, `go test -race ./pkg/kube/client/...` green):
  - `TestKubeForProviderConfigCaching` - same credentials reuse the client, different kubeconfig or different identity credentials build a new one, LRU eviction with a bound of 1, a negative bound caches without limit (the LRU would otherwise evict every entry on insert).
  - `TestKubeForProviderConfigConcurrentBuildsCollapse` - 16 concurrent cold-cache callers share one client and trigger one build.
  - `TestKubeForProviderConfigOutlivesContext` - hermetic Google `external_account` flow against an httptest STS and API server; the build context is cancelled and a request through the cached REST config still carries the exchanged bearer token. Fails with `Post ...: context canceled` without the `WithoutCancel`.
  - `TestKubeForProviderConfigRESTConfig` - `QPS == -1`, returned config is a copy.
  - `TestDigestWrite` - field boundaries and embedded delimiters do not alias.
  - `TestBearerAuthRoundTripper` - the EKS token source is called with the request context and the token lands in `Authorization`.
- Benchmarked on kind as described above against `main` and this branch with a Go reconcile load test (to be proposed separately); raw output available on request.
- Identity wrapper behavior under caching was confirmed at source for `gke`, `azure`, `upbound`, `aws` and `nebius` (self-refreshing token sources behind the cached transports, context handling per wrapper); the kubeconfig and in-cluster paths were exercised live, the cloud identity paths were not.

### DIFF
--- a/cmd/provider/main.go
+++ b/cmd/provider/main.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"math"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -106,6 +107,7 @@ func main() {
 		// Additional legacy field managers to upgrade to Server-side apply field manager
 		legacyCSAFieldManagers = app.Flag("legacy-csa-field-managers", "Additional legacy client-side apply Kubernetes field manager names for upgrading to SSA field manager").Default().Strings()
 	)
+	app.Validate(func(*kingpin.Application) error { return validateClientCacheSize(*clientCacheSize) })
 	kingpin.MustParse(app.Parse(os.Args[1:]))
 
 	zl := zap.New(zap.UseDevMode(*debug), UseISO8601())
@@ -312,4 +314,13 @@ func canWatchCRD(ctx context.Context, mgr manager.Manager) (bool, error) {
 		}
 	}
 	return true, nil
+}
+
+// validateClientCacheSize rejects flag values that do not fit the builder's
+// int bound instead of letting the conversion wrap around.
+func validateClientCacheSize(size uint) error {
+	if size > math.MaxInt {
+		return errors.Errorf("--client-cache-size must not exceed %d", math.MaxInt)
+	}
+	return nil
 }
--- a/cmd/provider/main.go
+++ b/cmd/provider/main.go
@@ -23,6 +23,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strconv"
 	"time"
 
 	"github.com/alecthomas/kingpin/v2"
@@ -62,6 +63,7 @@ import (
 	controllerNamespaced "github.com/crossplane-contrib/provider-kubernetes/internal/controller/namespaced"
 	"github.com/crossplane-contrib/provider-kubernetes/internal/features"
 	"github.com/crossplane-contrib/provider-kubernetes/internal/version"
+	kubeclient "github.com/crossplane-contrib/provider-kubernetes/pkg/kube/client"
 
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 )
@@ -89,6 +91,7 @@ func main() {
 		pollJitterPercentage    = app.Flag("poll-jitter-percentage", "Percentage of jitter to apply to poll interval. It cannot be negative, and must be less than 100.").Default("10").Uint()
 		leaderElection          = app.Flag("leader-election", "Use leader election for the controller manager.").Short('l').Default("false").Envar("LEADER_ELECTION").Bool()
 		maxReconcileRate        = app.Flag("max-reconcile-rate", "The number of concurrent reconciliations that may be running at one time.").Default("100").Int()
+		clientCacheSize         = app.Flag("client-cache-size", "Maximum number of cached target cluster clients, one per distinct ProviderConfig credential set; the least recently used client is evicted beyond this bound. 0 caches without bound.").Default(strconv.Itoa(kubeclient.DefaultClientCacheSize)).Envar("CLIENT_CACHE_SIZE").Uint()
 		sanitizeSecrets         = app.Flag("sanitize-secrets", "when enabled, redacts Secret data from Object status").Default("false").Envar("SANITIZE_SECRETS").Bool()
 		webhookPort             = app.Flag("webhook-port", "The port the webhook server listens on.").Default("9443").Envar("WEBHOOK_PORT").Int()
 		metricsBindAddress      = app.Flag("metrics-bind-address", "The address the metrics server listens on").Default(":8080").Envar("METRICS_BIND_ADDRESS").String()
@@ -200,6 +203,9 @@ func main() {
 		RemoveManagedFields:  *removeManagedFields,
 		PollJitter:           pollJitter,
 		PollJitterPercentage: *pollJitterPercentage,
+		ClientBuilder: kubeclient.NewIdentityAwareBuilder(mgr.GetClient(),
+			kubeclient.WithLogger(log),
+			kubeclient.WithClientCacheSize(int(*clientCacheSize))),
 	}
 
 	if *enableManagementPolicies {
--- a/internal/controller/namespaced/object/object.go
+++ b/internal/controller/namespaced/object/object.go
@@ -154,7 +154,7 @@ type ResourceSyncer interface {
 }
 
 // Setup adds a controller that reconciles Object managed resources.
-func Setup(mgr ctrl.Manager, o controller.Options, sanitizeSecrets bool, removeManagedFields bool, pollJitterPercentage uint, legacyCSAFieldManagers []string) error { // nolint:gocyclo // Too many branches due to alpha features, hopefully we can clean them up after we graduate them.
+func Setup(mgr ctrl.Manager, o controller.Options, clientBuilder kubeclient.Builder, sanitizeSecrets bool, removeManagedFields bool, pollJitterPercentage uint, legacyCSAFieldManagers []string) error { // nolint:gocyclo // Too many branches due to alpha features, hopefully we can clean them up after we graduate them.
 	name := managed.ControllerName(v1alpha1.ObjectGroupKind)
 	l := o.Logger.WithValues("controller", name)
 
@@ -183,7 +183,7 @@ func Setup(mgr ctrl.Manager, o controller.Options, sanitizeSecrets bool, removeM
 		removeManagedFields: removeManagedFields,
 		kube:                mgr.GetClient(),
 		usage:               resource.NewProviderConfigUsageTracker(mgr.GetClient(), &apisv1alpha1.ProviderConfigUsage{}),
-		clientBuilder:       kubeclient.NewIdentityAwareBuilder(mgr.GetClient()),
+		clientBuilder:       clientBuilder,
 	}
 
 	if o.Features.Enabled(features.EnableBetaServerSideApply) {
@@ -258,9 +258,9 @@ func Setup(mgr ctrl.Manager, o controller.Options, sanitizeSecrets bool, removeM
 
 // SetupGated registers a gated controller that reconciles Object managed resources.
 // The controller setup is initiated after the CRD for Object becomes available.
-func SetupGated(mgr ctrl.Manager, o controller.Options, sanitizeSecrets bool, removeManagedFields bool, pollJitterPercentage uint, legacyCSAFieldManagers []string) error {
+func SetupGated(mgr ctrl.Manager, o controller.Options, clientBuilder kubeclient.Builder, sanitizeSecrets bool, removeManagedFields bool, pollJitterPercentage uint, legacyCSAFieldManagers []string) error {
 	o.Gate.Register(func() {
-		if err := Setup(mgr, o, sanitizeSecrets, removeManagedFields, pollJitterPercentage, legacyCSAFieldManagers); err != nil {
+		if err := Setup(mgr, o, clientBuilder, sanitizeSecrets, removeManagedFields, pollJitterPercentage, legacyCSAFieldManagers); err != nil {
 			mgr.GetLogger().Error(err, "unable to setup reconciler", "gvk", v1alpha1.ObjectGroupVersionKind.String())
 		}
 	}, v1alpha1.ObjectGroupVersionKind)
--- a/internal/controller/namespaced/object/object.go
+++ b/internal/controller/namespaced/object/object.go
@@ -77,6 +77,7 @@ const (
 )
 
 const (
+	errNilClientBuilder  = "a client builder is required"
 	errGetProviderConfig = "cannot get provider config"
 	errTrackPCUsage      = "cannot track ProviderConfig usage"
 	errGetObject         = "cannot get object"
@@ -155,6 +156,9 @@ type ResourceSyncer interface {
 
 // Setup adds a controller that reconciles Object managed resources.
 func Setup(mgr ctrl.Manager, o controller.Options, clientBuilder kubeclient.Builder, sanitizeSecrets bool, removeManagedFields bool, pollJitterPercentage uint, legacyCSAFieldManagers []string) error { // nolint:gocyclo // Too many branches due to alpha features, hopefully we can clean them up after we graduate them.
+	if clientBuilder == nil {
+		return errors.New(errNilClientBuilder)
+	}
 	name := managed.ControllerName(v1alpha1.ObjectGroupKind)
 	l := o.Logger.WithValues("controller", name)
 
@@ -259,6 +263,9 @@ func Setup(mgr ctrl.Manager, o controller.Options, clientBuilder kubeclient.Buil
 // SetupGated registers a gated controller that reconciles Object managed resources.
 // The controller setup is initiated after the CRD for Object becomes available.
 func SetupGated(mgr ctrl.Manager, o controller.Options, clientBuilder kubeclient.Builder, sanitizeSecrets bool, removeManagedFields bool, pollJitterPercentage uint, legacyCSAFieldManagers []string) error {
+	if clientBuilder == nil {
+		return errors.New(errNilClientBuilder)
+	}
 	o.Gate.Register(func() {
 		if err := Setup(mgr, o, clientBuilder, sanitizeSecrets, removeManagedFields, pollJitterPercentage, legacyCSAFieldManagers); err != nil {
 			mgr.GetLogger().Error(err, "unable to setup reconciler", "gvk", v1alpha1.ObjectGroupVersionKind.String())
--- a/pkg/kube/client/client.go
+++ b/pkg/kube/client/client.go
@@ -15,16 +15,23 @@ package client
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/binary"
+	"encoding/hex"
+	"hash"
 	"net/http"
 	"net/url"
 	"strings"
 
 	"github.com/pkg/errors"
+	"golang.org/x/sync/singleflight"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/clientcmd/api"
+	"k8s.io/utils/lru"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/crossplane/crossplane-runtime/v2/pkg/logging"
 	"github.com/crossplane/crossplane-runtime/v2/pkg/resource"
 	xpv2 "github.com/crossplane/crossplane/apis/v2/core/v2"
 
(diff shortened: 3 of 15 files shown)


cmd/provider/main.go
```diff
@@ -89,6 +91,7 @@ func main() {
 		pollJitterPercentage    = app.Flag("poll-jitter-percentage", "Percentage of jitter to apply to poll interval. It cannot be negative, and must be less than 100.").Default("10").Uint()
 		leaderElection          = app.Flag("leader-election", "Use leader election for the controller manager.").Short('l').Default("false").Envar("LEADER_ELECTION").Bool()
 		maxReconcileRate        = app.Flag("max-reconcile-rate", "The number of concurrent reconciliations that may be running at one time.").Default("100").Int()
+		clientCacheSize         = app.Flag("client-cache-size", "Maximum number of cached target cluster clients, one per distinct ProviderConfig credential set; the least recently used client is evicted beyond this bound. 0 caches without bound.").Default(strconv.Itoa(kubeclient.DefaultClientCacheSize)).Envar("CLIENT_CACHE_SIZE").Uint()
```
bobh66, line 95: I don't think relying on the user to configure the cache size is practical - the information required to determine this number may not be available and could vary significantly over time.  The cost of guessing wrong or leaving the default is a significant performance hit that may be hard to debug.  Ideally the cache size would be dynamically configured based on the number of ProviderConfigs, if the client builder is able to modify the cache size.  Otherwise I think the default value needs to be much larger - maybe 50 or 100 - to ensure that the cache does not overflow and trigger the performance problems.

jonasz-lasut: I'll check if we can build the cache size dynamically, that's a good proposal but I don't think it should be equivalent to ProviderConfigs, as we cache based on the ProviderConfig identity/credentials and not on the ProviderConfig's name/namespace so does not make sense to build 1:1 relation

On the default value being much larger, may I ask you why do you think that should be the case? Cache is build from the credentials and identity (if present) and not from the number of ProviderConfigs/ClusterProviderConfigs, do we expect the most common use-case for provider-kubernetes to be one controlplane cluster that manages Objects in >8 workload clusters? I don't mind setting the value to a higher number (let's say 128) but that will lead to old clients not getting removed from cache quickly if for example the kubeconfig in providerconfig changes some field during rotation and old-one is now unusable

jonasz-lasut: I think that the best approach if we want to change this behavior would be to increase the default but add a Idle-TTL eviction so that we delete the rotated-out credentials, however I still think that default of 8 with how the cache keys are built and a flag to increase it is enough as long as we add documentation on how to tweak it

bobh66: That's probably a reasonable solution, but I am concerned that 8 is a pretty small number.  We have some controlplane clusters that manage 1 workload cluster each, and others that control 20+ workload clusters, with plans to double or triple that number.  I'm not sure there is a good way to determine a common configuration.

Documentation will be key - and maybe a "breaking change" label is appropriate since the previous behavior was unlimited and now it is being significantly reduced.  Maybe most people won't care, but they should have the opportunity to proactively decide they don't care.

Does the cache size affect memory usage before the cache entries are used?  If the memory is not allocated until the cache entry is created, and if it is reaped when the entry is deleted, then a larger cache could have no impact, especially if a TTL is used to clean old entries.

jonasz-lasut: The cache size does not impact memory in any substantial manner before the entries are used.
I've also evaluated the TTL part and we'd have to implement it using sync maps or smth similar, so unfortunately we can't reuse the existing cache mechanism from upstream k8s.

I'd probably stay with LRU, maybe little bigger default and for sure a proper documentation.

I'd also just like to add that `"breaking change" label is appropriate since the previous behavior was unlimited and now it is being significantly reduced.` this part is not 100% correct. We have never cached the clients and instead created them per reconciliation. What that means with current cache in mind - if you have more unique provider credentials than the cache size you'd have to create a new client for a reconciliation possibly rotating out client that was supposed to used in next queued reconciliation.
In worst case scenario you end up with every client being created fresh, which is the same behavior as prior to this change so it's not breaking but you may also lose the cache benefits if it's misconfigured

bobh66: Maybe I'm confused - my understanding is that the currently-released versions don't have the API discovery issue from the latest controller-runtime, which they would have in the latest main branch without this change.  Overflowing a small cache would presumably trigger the performance issues to some degree as API discovery has to be run more often than necessary.  I'm ok with not setting breaking change, but if there is the possibility that a too-small cache will trigger performance problems that we don't have on the current releases, it definitely needs to be called out at the same urgency as a breaking change.

jonasz-lasut: Unfortunately [v1.3.0](https://github.com/crossplane-contrib/provider-kubernetes/releases/tag/v1.3.0) already has the performance issues as it was running on controller-runtime > 0.20

I'm happy to set the default to a higher number but I'd rather not set it to a value in 50-100 range as without Idle-TTL eviction in most cases it will just grow without any visible benefit during credentials rotation etc.

And I **100% agree** that we need to mark this update with a big red warning banner during the release to make sure that people take conscious decision when setting the cache value and they're aware of the change

bobh66: 8 is probably as good a number as any.  The debug message on eviction should help to diagnose any performance issues, and the introduction of the cache will hopefully alleviate the issues in 1.3.0.  Thanks!

jonasz-lasut: I'll create a small CLI that you can run against the cluster to get an information about the smallest required cache size to not hit constant evictions, that should help the community to properly adopt this feature. Should be straightforward as cache key builder can be imported out of this provider.

cmd/provider/main.go
```diff
@@ -200,6 +203,9 @@ func main() {
 		RemoveManagedFields:  *removeManagedFields,
 		PollJitter:           pollJitter,
 		PollJitterPercentage: *pollJitterPercentage,
+		ClientBuilder: kubeclient.NewIdentityAwareBuilder(mgr.GetClient(),
+			kubeclient.WithLogger(log),
+			kubeclient.WithClientCacheSize(int(*clientCacheSize))),
```
erhancagirici, line 210: very nit: this has an overflow potential resulting in an unbounded cache

pkg/kube/client/client.go
```diff
@@ -66,39 +73,131 @@ func (fn BuilderFn) KubeForProviderConfig(ctx context.Context, pc kconfig.Provid
...
+func WithLogger(l logging.Logger) BuilderOption {
+	return func(b *IdentityAwareBuilder) {
+		b.log = l
+	}
+}
+
+// WithClientCacheSize bounds the number of cached target cluster clients, one
+// per distinct credential set; the least recently used client is evicted
+// beyond the bound. A size of 0 removes the bound.
+func WithClientCacheSize(size int) BuilderOption {
```
erhancagirici, line 115: nit: negative values result in an unbounded cache, might be good to explicitly document

internal/controller/namespaced/object/object.go
```diff
@@ -183,7 +183,7 @@ func Setup(mgr ctrl.Manager, o controller.Options, sanitizeSecrets bool, removeM
 		removeManagedFields: removeManagedFields,
 		kube:                mgr.GetClient(),
 		usage:               resource.NewProviderConfigUsageTracker(mgr.GetClient(), &apisv1alpha1.ProviderConfigUsage{}),
-		clientBuilder:       kubeclient.NewIdentityAwareBuilder(mgr.GetClient()),
+		clientBuilder:       clientBuilder,
```
erhancagirici, line 190: nit: now this is passed down through `pcontroller.Options`, might be better do an explicit nil check and fail the setup. When nil, the reconciler setup passes, and this will nil-panic at the first reconcile.

Valid for other `Setup(Gated)` functions as well.

pkg/kube/client/client.go
```diff
@@ -66,39 +73,131 @@ func (fn BuilderFn) KubeForProviderConfig(ctx context.Context, pc kconfig.Provid
...
+	})
 	if err != nil {
-		return nil, nil, errors.Wrapf(err, "cannot create Kubernetes client for provider")
+		return nil, nil, errors.Wrap(err, "cannot create Kubernetes client for provider")
 	}
-	return k, rc, nil
+	cached := v.(cachedClient)
+	// The config is shared by every caller of this credential set; hand out a
+	// copy so that nobody can mutate it underneath the cached client.
+	return cached.kube, rest.CopyConfig(cached.rc), nil
```
erhancagirici, line 182: note: this is a shallow copy (OK for our purposes)

pkg/kube/client/client.go
```diff
@@ -127,21 +228,36 @@ func (b *IdentityAwareBuilder) restForProviderConfig(ctx context.Context, pc kco
 		}
 	}
 
+	// The client built from this config is shared by every reconcile of the
+	// same credential set, so a per-client token bucket (client-go defaults
+	// to 5 QPS with a burst of 10 when none is set) would serialize them.
+	// Concurrency is bounded by --max-reconcile-rate and the API server's
+	// priority and fairness instead.
+	rc.QPS = -1
```
erhancagirici, line 236: note: this will also "un-throttle" everything else the client handles, like long-living informers (watch-enabled MRs) and SSA discovery client etc.
We'll solely rely on APF of the target cluster, and possibly the practical effect is insignificant, but this is a behavior change ,IMHO, we should document.

pkg/kube/client/client.go
```diff
@@ -127,21 +228,36 @@ func (b *IdentityAwareBuilder) restForProviderConfig(ctx context.Context, pc kco
...
 				}
 			default:
 				creds, err := resource.CommonCredentialExtractor(ctx, id.Source, b.local, id.CommonCredentialSelectors)
 				if err != nil {
 					return nil, errors.Wrap(err, errExtractGoogleCredentials)
 				}
+				digestWrite(digest, creds)
 
-				if err := gke.WrapRESTConfig(ctx, rc, creds, gke.DefaultScopes...); err != nil {
+				if err := gke.WrapRESTConfig(wrapCtx, rc, creds, gke.DefaultScopes...); err != nil {
```
erhancagirici, line 262: this is the only part that I am a bit concerned. `wrapCtx` resolves the per-reconcile cancellation, but this time it becomes fully unbounded.

In particular for GKE case, at pkg/kube/client/gke/gke.go:44

```go
creds, err := google.CredentialsFromJSON(ctx, credentials, scopes...)
  ts = creds.TokenSource
  rc.Wrap(func(rt http.RoundTripper) http.RoundTripper {
      return &oauth2.Transport{Source: ts, Base: rt}
  })
```
AFAIU, the `google.CredentialsFromJSON` is meant to be one-use only.
The returned token source stores the given ctx in its implementation (due to oauth.TokenSource.Token() interface has no context in its signature). So, requests will have no cancellation.

Other aspect to carefully consider is: the returned token source impl holds a mutex. 
```go
 func (s *reuseTokenSource) Token() (*Token, error) {
        s.mu.Lock()
        defer s.mu.Unlock()
        if s.t.Valid() { return s.t, nil }
        t, err := s.new.Token()   // network call with held lock
        ...
  }
```
So the `ts` instance and its mutex is shared among concurrent reconciles of the same credential set. The long-living informer caches and discovery clients also utilize the same token source. One hiccup in token retrieval, blocks the whole set of reconciles relying on the same cred set.

atm I have no quick solution to propose, but one possibility to explore:
- revisit deprecated `google.CredentialsFromJSON` (we should do regardless) and see if there is a context-aware token source impl.
- possibly ditch oauth2.Transport, and implement a thin RoundTripper.

erhancagirici (commented):
@jonasz-lasut thanks! LGTM already, with few nits

pkg/kube/client/client.go
jonasz-lasut: negative values result in always empty cache in k8s.io/utils' LRU cache implementation
I think we should normalize the negatives to zero and document like you've said

cmd/provider/main.go
jonasz-lasut: Added `validateClientCacheSize` kingpin validator to ensure that size is smaller than math.MaxInt

internal/controller/namespaced/object/object.go
jonasz-lasut: Added `errNilClientBuilder` error to all places in Setup/SetupGated where we may end up with a nil client builder

pkg/kube/client/client.go
jonasz-lasut: added normalization for consumers of this pkg, cache size is build as `max(size, 0)` so negative values correctly mean unbound cache if the value is passed directly when this pkg is used as library

when this provider is used as a standalone provider negative values are impossible to pass as flag because we parse the --client-cache-size value as Uint, which errors on negative values

pkg/kube/client/client.go
jonasz-lasut: Added entry in readme:

`Client-side rate limiting is disabled for these clients...`

pkg/kube/client/client.go
jonasz-lasut: that was intended as transport wrappers and TLS material are the cached client's identity.

pkg/kube/client/client.go
jonasz-lasut: Please check this [commit](https://github.com/crossplane-contrib/provider-kubernetes/pull/543/changes/c85058ab8f2bf1fad37ec48abcd00e82b33d4241) 

gke.WrapRESTConfig installs an http.Client with a 30 s timeout under oauth2.HTTPClient in the context it hands to the Google credentials. Every oauth2 token source resolves its HTTP client from that key (oauth2.NewClient(ctx, nil) then internal.ContextClient), and context.WithoutCancel keeps values, so the exchange behind the cached transport is bounded even though it no longer inherits the reconcile deadline. A silent token endpoint costs one 30 s stall per token lifetime instead of a hang. Covered by TestWrapRESTConfigTokenFetch with an httptest STS that never answers.

The shared reuseTokenSource mutex stays, as you noted: oauth2.TokenSource.Token() has no context, so concurrent requests of a credential set still wait for the one refresh in flight, now for at most 30 s. The context-aware option is cloud.google.com/go/auth, whose Credentials.Token(ctx) takes the request context and refreshes stale tokens in the background without blocking callers. That also retires the deprecated google.CredentialsFromJSON. I would do it as its own PR since it adds a dependency and rewrites the wrapper. I'll open an issue for that after we agree on moving it to a separate effort

cmd/provider/main.go
erhancagirici: my summary & 2c, we should follow up on this.

while I ack that there is no good default static value for the cache capacity, IMHO, the flag exposes a low-level impl detail, and I see it as an escape hatch only. 
(In general) For most provider consumers, I believe the flag would be mostly obscure (e.g. provider installed as a dependency etc.)
For the provider admin persona, it still needs to be evaluated and not something set-and-forget, potentially dynamic updates needed depending on the usage pattern. 

I would also have suggested to go for a higher default, but AFAIU, size of an individual cache entry can vary a lot and might be significant(especially depending on the target cluster's discovery docs etc). So overshooting cache size would waste memory for clusters with low cardinality creds, because rotations keeping the stale entries longer.

Number of providerconfigs is a good upper bound for active creds, with today's PC spec. Though AFAIU the edge case is when different PCs result in the equivalent creds/kubeconfig, overshooting the optimal size. 

TLDR, this PR makes the current situation better in any case, but let's evaluate a stricter cache eviction as a follow-up

erhancagirici (approved):
thanks @jonasz-lasut, LGTM and approved. Few separate follow-ups as noted:

- replace deprecated`google.CredentialsFromJSON` in GKE cred resolution
- possibly autopilot cache size or stricter eviction: see https://github.com/crossplane-contrib/provider-kubernetes/pull/543/changes#r3913268493
- cache eviction metrics: if we cannot autopilot cache size (e.g. expect consumer to act), we should provide a stable interface apart from logs

jonasz-lasut:
Thank you @erhancagirici I'll open the issues for:
* deprecated google.CredentialsFromJSON in GKE replacement
* autopilot cache size (the cache size has to stay static but I think we could do a ProviderConfig/ClusterProviderConfig list call on startup and figure out the number of unique credentials that are present to build the cache size) and metrics related to cache

PR #556, opened 2026-08-27
Fix ObservedObjectCollection failing to create Objects due to incorrectly serialized deletionPropagationPolicy

### LINKED ISSUE #555: regression: ObservedObjectCollection fails to create Objects due to incorrect deletionPropagationPolicy value in v1.3.0
First reported by Israel Hakimi in community slack channel, I've copy-pasted his report:

### What happened?
Provider version: xpkg.upbound.io/crossplane-contrib/provider-kubernetes:v1.3.0
Affected API: observedobjectcollections.kubernetes.m.crossplane.io/v1alpha1 (namespaced)
Crossplane runtime: crossplane-runtime/v2 v2.3.3

Summary

v1.3.0 added a new deletionPropagationPolicy field to the objects.kubernetes.m.crossplane.io CRD (enum Orphan/Background/Foreground, default Background, no empty string allowed). The ObservedObjectCollection controller (internal/controller/namespaced/observedobjectcollection/reconciler.go, observedObjectPatch func, ~line 206-217) was not updated for this new field: when it builds the child Object it creates per matched resource, it only sets ManagedResourceSpec and ForProvider.Manifest:
```go
observedObject := &objectv1alpha1.Object{
    ObjectMeta: metav1.ObjectMeta{
        Name:      name,
        Namespace: collection.GetNamespace(),
        OwnerReferences: []metav1.OwnerReference{...},
    },
    Spec: objectv1alpha1.ObjectSpec{
        ManagedResourceSpec: xpv2.ManagedResourceSpec{
            ProviderConfigReference: &collection.Spec.ProviderConfigReference,
            ManagementPolicies:      []xpv2.ManagementAction{xpv2.ManagementActionObserve},
        },
        ForProvider: objectv1alpha1.ObjectParameters{
            Manifest: runtime.RawExtension{Raw: []byte(manifest)},
        },
    },
}
```
DeletionPropagationPolicy is left at its Go zero value (""). Because the field is a plain string (not a pointer with omitempty), the resulting unstructured payload sent to the API server includes deletionPropagationPolicy: "" explicitly. CRD default-value injection only applies to absent fields, so the explicit empty string hits the new enum validation and the create is rejected — every reconcile, forever (tight error/retry loop).

### How can we reproduce it?
Install provider-kubernetes:v1.3.0.
Apply an ObservedObjectCollection in any namespace, e.g.:
```yaml
apiVersion: kubernetes.m.crossplane.io/v1alpha1
kind: ObservedObjectCollection
metadata:
  name: endpointslice-observer-example
  namespace: resiliency
spec:
  providerConfigRef:
    name: kubernetes-cluster-dev-00
    kind: ProviderConfig
  observeObjects:
    apiVersion: discovery.k8s.io/v1
    kind: EndpointSlice
    namespace: resiliency
    selector:
      matchLabels:
        kubernetes.io/service-name: istio-test-nginx
```
Any matching EndpointSlice (or other target kind) triggers creation of a child Object.
Expected: child Object is created successfully, observing the matched resource.

Actual: create is rejected every time with:
cannot create observed object: Object.kubernetes.m.crossplane.io "endpointslice-observer-example-e5f111e" is invalid:
[spec.forProvider.deletionPropagationPolicy: Unsupported value: "": supported values: "Orphan", "Background", "Foreground",
<nil>: Invalid value: null: some validation rules were not checked because the object was invalid; correct the existing
errors to complete validation]
ObservedObjectCollection.status.conditions shows Synced: False, reason: ReconcileError permanently, and any composite resource depending on it (e.g. via extraResources) is blocked from reaching Ready.

Suggested fix: in observedObjectPatch, either omit DeletionPropagationPolicy via a pointer type with omitempty, or explicitly set it to "Background" (or "Orphan", appropriate for observe-only resources) when constructing ForProvider.
Workaround: pin to v1.2.1 (last release before deletionPropagationPolicy was added) until fixed.`

jonasz-lasut:

Fixes #555
* adds omitempty to deletionPropagationPolicy so that it does not get serialized to empty string `""`
* adds unit tests to cover regression
* updates the `ObservedObjectCollection` example to be uptestable -configmaps did not have status.conditions, which resulted in e2e tests always failing. Updated `ObservedObjectCollection` to observe Deployments in crossplane-system namespace, which always exists in e2e tests and contains crossplane & crossplane-rbac-manager deployments.
* adds e2e target for `ObservedObjectCollection`

I have:

Added E2E target to test ObservedObjectCollection, added unit tests

### DIFF
--- a/examples/cluster/collection/collection.yaml
+++ b/examples/cluster/collection/collection.yaml
@@ -1,24 +1,28 @@
 ---
 apiVersion: v1
-kind: ConfigMap
+kind: Pod
 metadata:
-  name: configmap-bar
-  namespace: default
   labels:
     foo: bar
-data:
-  sample-key: "sample-value"
+  name: pod-bar
+  namespace: crossplane-system
+spec:
+  containers:
+  - image: registry.k8s.io/pause
+    name: pod-bar
 
 ---
 apiVersion: v1
-kind: ConfigMap
+kind: Pod
 metadata:
-  name: configmap-foo
-  namespace: default
   labels:
     foo: bar
-data:
-  sample-key: "sample-value2"
+  name: pod-foo
+  namespace: crossplane-system
+spec:
+  containers:
+  - image: registry.k8s.io/pause
+    name: pod-foo
 ---
 apiVersion: kubernetes.crossplane.io/v1alpha1
 kind: ObservedObjectCollection
@@ -27,7 +31,7 @@ metadata:
 spec:
   observeObjects:
     apiVersion: v1
-    kind: ConfigMap
+    kind: Pod
     selector:
       matchLabels:
         foo: bar
@@ -39,4 +43,3 @@ spec:
         a1: v1
   providerConfigRef:
     name: kubernetes-provider
-
--- a/examples/namespaced/collection/collection.yaml
+++ b/examples/namespaced/collection/collection.yaml
@@ -1,24 +1,29 @@
 ---
 apiVersion: v1
-kind: ConfigMap
+kind: Pod
 metadata:
-  name: configmap-bar
-  namespace: other-ns
   labels:
     foo: bar
-data:
-  sample-key: "sample-value"
+  name: pod-bar
+  namespace: crossplane-system
+spec:
+  containers:
+  - image: registry.k8s.io/pause
+    name: pod-bar
 
 ---
 apiVersion: v1
-kind: ConfigMap
+kind: Pod
 metadata:
-  name: configmap-foo
-  namespace: default
   labels:
     foo: bar
-data:
-  sample-key: "sample-value2"
+  name: pod-foo
+  namespace: crossplane-system
+spec:
+  containers:
+  - image: registry.k8s.io/pause
+    name: pod-foo
+
 ---
 apiVersion: kubernetes.m.crossplane.io/v1alpha1
 kind: ObservedObjectCollection
@@ -28,7 +33,7 @@ metadata:
 spec:
   observeObjects:
     apiVersion: v1
-    kind: ConfigMap
+    kind: Pod
     selector:
       matchLabels:
         foo: bar
@@ -41,4 +46,3 @@ spec:
   providerConfigRef:
     kind: ClusterProviderConfig
     name: kubernetes-provider
-
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ CROSSPLANE_NAMESPACE = crossplane-system
 -include build/makelib/local.xpkg.mk
 -include build/makelib/controlplane.mk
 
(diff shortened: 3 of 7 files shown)


examples/cluster/collection/collection.yaml
```diff
@@ -1,36 +1,14 @@
----
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: configmap-bar
-  namespace: default
-  labels:
-    foo: bar
```
erhancagirici, line 8: First of all, thanks for fixing the example 🙌 

nit: in uptest, it is possible to customize per-manifest the condition to check by setting the annotation `uptest.upbound.io/conditions`
In this particular case of ConfigMap,  I think setting an empty string value would make them uptestable.

examples/cluster/collection/collection.yaml
```diff
@@ -1,36 +1,14 @@
...
 spec:
   observeObjects:
-    apiVersion: v1
-    kind: ConfigMap
+    apiVersion: apps/v1
+    kind: Deployment
     selector:
       matchLabels:
-        foo: bar
+        app.kubernetes.io/name: crossplane
```
erhancagirici: very nit: My very small concern is the stability of `app.kubernetes.io/name: crossplane` annotation, and the behavior of the empty match.

AFAIU, by skimming the controller code, empty match results in a ready `ObservedObjectCollection` MR. (I know the chances are quite low but) So, if somehow this label changes during xp bumps etc, this MR example will silently succeed in uptest. 

If possible, I'd prefer to keep the original ConfigMap-based example (by making it uptestable per my other comment) as well, and add this one as an extra. Original example is relatively more "self-contained" in terms of dependencies.

examples/namespaced/collection/collection.yaml
```diff
@@ -1,37 +1,15 @@
...
 spec:
   observeObjects:
-    apiVersion: v1
-    kind: ConfigMap
+    apiVersion: apps/v1
+    kind: Deployment
     selector:
       matchLabels:
-        foo: bar
+        app.kubernetes.io/name: crossplane
```
erhancagirici: nit: same as cluster-scoped https://github.com/crossplane-contrib/provider-kubernetes/pull/556/changes#r3872877494

erhancagirici (approved):
@jonasz-lasut thanks for the fix! LGTM already, just left a very small nit for e2e tests.

examples/cluster/collection/collection.yaml
jonasz-lasut: let me test that, if conditons: "" solves this issue than I'm more than happy to restore it

examples/cluster/collection/collection.yaml
jonasz-lasut: Let me reply here to make the discussion easy to follow. Unfortunately `uptest.upbound.io/conditions: ""` gets templated to an exact match on condition of type `""` which is never satisfied. However I was not happy with the crossplane deployment solution so I've decided to just create two pods instead of configmaps in the e2e example using the `pause` image as it's extremely light (300kB) and starts in few seconds at most

erhancagirici (approved):

PR #560, opened 2026-09-02
Add a client cache autopilot as a replacement for user-provided flag

### LINKED ISSUE #559: client cache: size the LRU from the cluster's credential sets at startup, and expose cache metrics
## Context

#543 (merged `9a1e83cb`, unreleased as of `v1.3.1`) bounds `IdentityAwareBuilder`'s per-credential-set client cache with `--client-cache-size` ([`cmd/provider/main.go:100`](https://github.com/crossplane-contrib/provider-kubernetes/blob/9a1e83cb2dc312df2dc08e6af033a598222e41eb/cmd/provider/main.go#L100)), default [`DefaultClientCacheSize = 8`](https://github.com/crossplane-contrib/provider-kubernetes/blob/9a1e83cb2dc312df2dc08e6af033a598222e41eb/pkg/kube/client/client.go#L78), LRU-evicted. The review ([r3813958630](https://github.com/crossplane-contrib/provider-kubernetes/pull/543#discussion_r3813958630) → [r3913268493](https://github.com/crossplane-contrib/provider-kubernetes/pull/543#discussion_r3913268493)) agreed the flag is an escape hatch rather than something a provider admin can reason about, and that a follow-up should make the size come from the cluster instead of from a guess. This issue tracks that follow-up, in two parts.

## Why a static default cannot be right for everyone

- **A bound that is too small is silent and expensive.** The overflow set rebuilds a client and re-runs aggregated discovery on every reconcile - exactly the #541 regression - and the only signal is a `Debug` log on eviction ([`client.go:136-137`](https://github.com/crossplane-contrib/provider-kubernetes/blob/9a1e83cb2dc312df2dc08e6af033a598222e41eb/pkg/kube/client/client.go#L136-L137)). @bobh66 runs control planes managing 1 cluster each *and* 20+ with plans to double or triple ([r3814188755](https://github.com/crossplane-contrib/provider-kubernetes/pull/543#discussion_r3814188755)).
- **A bound that is too large keeps rotated-out clients alive.** **Measured (kind, 2026-08-19):** one primed controller-runtime client retains ~0.4 MiB on a 30-CRD cluster and **~6 MiB on a 1530-CRD cluster** (aggregated discovery into the lazy `RESTMapper`, cr ≥ v0.20). When a kubeconfig `Secret` rotates the old digest's entry stays until LRU pressure evicts it, so a large default trades a silent regression for silent memory growth on CRD-heavy fleets - the point @erhancagirici makes in [r3913268493](https://github.com/crossplane-contrib/provider-kubernetes/pull/543#discussion_r3913268493).
- **"Number of ProviderConfigs" is an upper bound, not the answer.** The key is the SHA-256 of the credential material + identity ([`restForProviderConfig`, `client.go:200`](https://github.com/crossplane-contrib/provider-kubernetes/blob/9a1e83cb2dc312df2dc08e6af033a598222e41eb/pkg/kube/client/client.go#L200)), not the PC name: N PCs sharing a kubeconfig are **one** entry, every `InjectedIdentity` PC collapses into **one** entry, and one PC whose `Secret` rotates is **several** over time.

## Proposal, part 1: size from the cluster at startup

The number is knowable, and a PoC already computes it: [jonasz-lasut/provider-k8s-cache-size](https://github.com/jonasz-lasut/provider-k8s-cache-size) (the CLI promised in [r3814745805](https://github.com/crossplane-contrib/provider-kubernetes/pull/543#discussion_r3814745805)) lists legacy `ProviderConfig` plus `ProviderConfig`/`ClusterProviderConfig` of `kubernetes.m.crossplane.io`, re-derives the builder's digest for each, groups them, and reports the minimum `--client-cache-size`; an oracle test pins its grouping to the real `NewIdentityAwareBuilder`. Move that logic in-tree, **generic**, because the cache is not ours alone: `pkg/kube/client` is imported by provider-helm, whose v1beta1 `ProviderConfig`/`ClusterProviderConfig` embed the same `kconfig.ProviderConfigSpec` ([`apis/cluster/v1beta1/types.go:43`](https://github.com/crossplane-contrib/provider-helm/blob/main/apis/cluster/v1beta1/types.go#L43), [`apis/namespaced/v1beta1/types.go:43`](https://github.com/crossplane-contrib/provider-helm/blob/main/apis/namespaced/v1beta1/types.go#L43)) and go through the same `KubeForProviderConfig(ctx, *pcSpec)` ([`release.go:213`](https://github.com/crossplane-contrib/provider-helm/blob/main/pkg/controller/cluster/release/release.go#L213)).

1. **Export the key.** `IdentityAwareBuilder` gains an exported per-spec cache-key accessor, so the sizing code derives keys through the builder's own digest rather than mirroring it (the PoC mirrors it today, and a field-order drift would go unnoticed).
2. **A generic calculator in `pkg/`**, e.g. `pkg/kube/client.ClientCacheSizeFor(ctx, specs []kconfig.ProviderConfigSpec) (int, error)`: derive the key for each spec, count distinct keys, apply **10 % proportional headroom**, floor at `DefaultClientCacheSize`. It knows nothing about which kinds the specs came from.
3. **Per-provider listing stays in each `cmd/provider/main.go`**: provider-kubernetes lists its three kinds, provider-helm lists its own, and each hands the specs to the calculator and the result to `WithClientCacheSize`. `--client-cache-size` stays as the explicit override (`0` = unbounded); when set, it wins.
4. **The size is static for the process lifetime** - `k8s.io/utils/lru` has no resize. A PC created later lands in the overflow set with today's degrade-to-per-reconcile behaviour until the next restart; the metrics below make that visible instead of silent.
5. **Startup list failure falls back, it does not fail startup.** RBAC denied, CRD not installed yet: log a warning and use `DefaultClientCacheSize` (or the flag). This is consistent with how the `Object` controllers already tolerate a missing `ClusterProviderConfig` CRD; the fallback path is the pre-this-change behaviour, nothing regresses.
6. **Retire the external CLI** once this lands - it was a PoC to make the number knowable, and the in-tree calculator supersedes it.

Known approximations, carried over from the PoC: `Environment`/`Filesystem` credential sources live in the pod, so their content is stood in for by the selector; a PC the provider itself would fail to build (missing `Secret`, `None` source) counts as +1 rather than being skipped, so the estimate errs high.

## Proposal, part 2: metrics

Put the collectors on the builder, in `pkg/kube/client`, so provider-helm gets them too; each provider registers them next to its existing recorders ([`main.go:210-211`](https://github.com/crossplane-contrib/provider-kubernetes/blob/9a1e83cb2dc312df2dc08e6af033a598222e41eb/cmd/provider/main.go#L210-L211)):

| series | type | source |
|---|---|---|
| `client_cache_size` | gauge | the bound in effect (computed, flag, or default) |
| `client_cache_entries` | gauge | `b.clients.Len()`, already logged at Debug on every build ([`client.go:173`](https://github.com/crossplane-contrib/provider-kubernetes/blob/9a1e83cb2dc312df2dc08e6af033a598222e41eb/pkg/kube/client/client.go#L173)) |
| `client_cache_events_total{event="hit\|miss\|evict"}` | counter | the eviction func at `client.go:136` plus the hit/miss branch in `KubeForProviderConfig` |

No per-credential or per-PC label - unbounded cardinality. A sustained `evict` rate is the signal that the bound is too small; that turns today's silent regression into an alert.

## Alternatives considered

| option | verdict |
|---|---|
| Raise the static default (50-100) | Rejected in review: without time-based eviction it only lengthens how long rotated-out ~6 MiB entries linger, and still guesses. |
| Idle-TTL eviction instead of a count bound | The right *companion* to a larger bound, but `k8s.io/utils/lru` has no TTL, so it means a custom store. Out of scope here; part 1 + 2 make the need measurable first. |
| Size 1:1 from the ProviderConfig count | Wrong by construction (see above): overshoots for shared kubeconfigs, and `InjectedIdentity` fleets would allocate N entries for one. |
| Compute the size inside `cmd/provider/main.go` only | Leaves provider-helm (and any other importer of `pkg/kube/client`) with the static default; the calculation is provider-agnostic once it takes specs, so it belongs in `pkg/`. |

## Reference

- #543 review: [r3813958630](https://github.com/crossplane-contrib/provider-kubernetes/pull/543#discussion_r3813958630), [r3814188755](https://github.com/crossplane-contrib/provider-kubernetes/pull/543#discussion_r3814188755), [r3814745805](https://github.com/crossplane-contrib/provider-kubernetes/pull/543#discussion_r3814745805), [r3913268493](https://github.com/crossplane-contrib/provider-kubernetes/pull/543#discussion_r3913268493)
- #541 - the per-reconcile discovery regression a too-small cache reintroduces
- [jonasz-lasut/provider-k8s-cache-size](https://github.com/jonasz-lasut/provider-k8s-cache-size) - the PoC to port and then retire

jonasz-lasut:
* The default bound of 8 cached target cluster clients cannot suit every
control plane: too small silently reverts to per-reconcile discovery
(#541), too large keeps rotated-out clients alive. Size it from the
cluster instead, and drop the --client-cache-size flag: the number is
knowable, so there is nothing left for an operator to guess.

  ClientCacheSizeFor in pkg/kube/client derives the cache key of every
  ProviderConfig spec through the builder, sixteen at a time, counts
  distinct credential sets, adds headroom of ten percent or four entries,
  whichever is larger, for rotating credentials and ProviderConfigs
  created later, and floors the result at DefaultClientCacheSize; a spec
  whose key cannot be derived counts as a set of its own. The calculator takes specs, so provider-helm
  can reuse it with its own kinds - [commit](https://github.com/crossplane-contrib/provider-kubernetes/pull/560/changes/edf52f0d8ed5365701bb490369b5e287a46a9620)
  
  At startup the provider lists its legacy ProviderConfig plus the
  namespaced ProviderConfig and ClusterProviderConfig kinds through a
  direct client (the manager client cannot read before the manager
  starts), within a 300 second budget, and logs the resulting size. A kind
  whose CRD is not installed contributes nothing; failing to build that
  client, to list the ProviderConfigs or to read their credentials within
  the budget fails startup: a partial count would not be a bound, so
  there is no fallback to the default (the issue proposed falling back
  with a warning).

* Split IdentityAwareBuilder's REST config construction into resolving a
ProviderConfig (read the credential material, digest it into the cache
key, build the REST config from the kubeconfig) and injecting the
identity (install the token sources), and run the injection only when a
client is actually built. A cache hit no longer runs the identity
wrappers, whose work (Google credential parsing, the AWS default
credential chain, Nebius SDK setup) was thrown away with the fresh REST
config on every reconcile. Export the key as ClientCacheKey so that sizing code can derive it
through the builder instead of mirroring the digest - [commit](https://github.com/crossplane-contrib/provider-kubernetes/pull/560/changes/8d43b3b7af34a0716447ec017db4903650f8dfa6)

* A too-small client cache degrades silently to per-reconcile discovery
(#541), with only a Debug log on eviction. Put collectors on
IdentityAwareBuilder so that every importer of pkg/kube/client can
register them next to its other recorders:

  \<ns\>_client_cache_size          gauge    the bound in effect (0 = unbounded)
  \<ns\>_client_cache_entries       gauge    clients currently cached
  \<ns\>_client_cache_events_total  counter  event="hit|miss|evict"

  The namespace is the embedding provider's own, as with upjet's metrics;
  provider-kubernetes registers them as provider_kubernetes_client_cache_*.
  No per-credential or per-ProviderConfig labels. A sustained eviction
  rate is the signal that the credential sets in use no longer fit the
  bound.
  
  Every event series is initialised to zero so that rates work from the
  first scrape. The entries gauge is refreshed after each Add rather than
  in the eviction callback, which the cache runs under its own lock - [commit](https://github.com/crossplane-contrib/provider-kubernetes/pull/560/changes/86d59b5fabf9be105ee2865577fc68ae3e32aa8e)

* Review follow-ups, one commit each: derive the keys concurrently and
raise the startup budget from 30 to 300 seconds - [commit](https://github.com/crossplane-contrib/provider-kubernetes/pull/560/changes/afb141dabadff65a262f6e9f18760900ca8adac7);
floor the headroom at four entries and describe the LRU churn on overflow
in the README (every new client evicts the least recently used one, whose
next reconcile rebuilds it) - [commit](https://github.com/crossplane-contrib/provider-kubernetes/pull/560/changes/9d45b5d95fa14cea07f5fc27cef384d01d17acba); reword the `miss` help
text, since concurrent misses of one credential set share a single build,
and ignore a nil metrics collector - [commit](https://github.com/crossplane-contrib/provider-kubernetes/pull/560/changes/4decd374ee2211347c5451c8f5305bd8bb67a631)

Fixes #559 

I have:

Run E2E tests, added unit tests, added log lines stating the computed cache size

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,12 @@ headroom for credentials that rotate, never below 8. The result is logged at
 startup. The provider does not start if its ProviderConfigs cannot be listed.
 
 The bound is fixed for the lifetime of the process, so ProviderConfigs created
-after startup share the headroom until the next restart.
+after startup share the headroom until the next restart. The cache reports
+`provider_kubernetes_client_cache_size` (the bound),
+`provider_kubernetes_client_cache_entries` (clients currently cached) and
+`provider_kubernetes_client_cache_events_total` with an `event` label of
+`hit`, `miss` or `evict`; a sustained eviction rate means the credential sets
+in use no longer fit the bound.
 
 Client-side rate limiting is disabled for these clients. A cached client is
 shared by every concurrent reconcile of its credential set, so client-go's
--- a/README.md
+++ b/README.md
@@ -31,11 +31,22 @@ spec:
 ## Target cluster clients
 
 The provider builds one client per distinct ProviderConfig credential set
-(kubeconfig plus identity) and keeps it in an LRU cache bounded by
-`--client-cache-size` (default 8, `0` removes the bound). Building a client is
+(kubeconfig plus identity) and keeps it in an LRU cache. Building a client is
 expensive because its REST mapper primes itself with full aggregated discovery
-on first use, so size the cache to the number of distinct credential sets the
-provider talks to.
+on first use, so the cache has to hold every credential set the provider talks
+to: a credential set that does not fit is rebuilt, discovery included, on every
+reconcile.
+
+At startup the provider sizes the cache from the cluster: it lists its
+`ProviderConfig` and `ClusterProviderConfig` objects, derives the cache key of
+each one the way the client builder does (so ProviderConfigs sharing a
+kubeconfig count once, and every in-cluster ProviderConfig counts as one), and
+bounds the cache at the number of distinct credential sets plus ten percent of
+headroom for credentials that rotate, never below 8. The result is logged at
+startup. The provider does not start if its ProviderConfigs cannot be listed.
+
+The bound is fixed for the lifetime of the process, so ProviderConfigs created
+after startup share the headroom until the next restart.
 
 Client-side rate limiting is disabled for these clients. A cached client is
 shared by every concurrent reconcile of its credential set, so client-go's
--- a/README.md
+++ b/README.md
@@ -34,17 +34,23 @@ The provider builds one client per distinct ProviderConfig credential set
 (kubeconfig plus identity) and keeps it in an LRU cache. Building a client is
 expensive because its REST mapper primes itself with full aggregated discovery
 on first use, so the cache has to hold every credential set the provider talks
-to: a credential set that does not fit is rebuilt, discovery included, on every
-reconcile.
+to. Once the credential sets in use outnumber the bound, every new client
+evicts the least recently used one, whose next reconcile rebuilds it (discovery
+included) and evicts another: the whole cache churns, not only the credential
+set that did not fit.
 
 At startup the provider sizes the cache from the cluster: it lists its
 `ProviderConfig` and `ClusterProviderConfig` objects, derives the cache key of
 each one the way the client builder does (so ProviderConfigs sharing a
 kubeconfig count once, and every in-cluster ProviderConfig counts as one), and
-bounds the cache at the number of distinct credential sets plus ten percent of
-headroom for credentials that rotate, never below 8. The result is logged at
+bounds the cache at the number of distinct credential sets plus headroom of
+ten percent or four entries, whichever is larger, for credentials that rotate
+and ProviderConfigs created later, never below 8. The result is logged at
 startup. The provider does not start if its ProviderConfigs and their
-credentials cannot be read within 300 seconds.
+credentials cannot be read within 300 seconds. When the provider runs outside a
+cluster, as in local development, ProviderConfigs with an in-cluster identity
+cannot be resolved and each counts as a credential set of its own, so the
+cache is merely oversized.
 
 The bound is fixed for the lifetime of the process, so ProviderConfigs created
 after startup share the headroom until the next restart. The cache reports
@@ -52,7 +58,7 @@ after startup share the headroom until the next restart. The cache reports
 `provider_kubernetes_client_cache_entries` (clients currently cached) and
 `provider_kubernetes_client_cache_events_total` with an `event` label of
 `hit`, `miss` or `evict`; a sustained eviction rate means the credential sets
-in use no longer fit the bound.
+in use no longer fit the bound and the cache is churning.
 
 Client-side rate limiting is disabled for these clients. A cached client is
 shared by every concurrent reconcile of its credential set, so client-go's
--- a/README.md
+++ b/README.md
@@ -43,7 +43,8 @@ each one the way the client builder does (so ProviderConfigs sharing a
 kubeconfig count once, and every in-cluster ProviderConfig counts as one), and
 bounds the cache at the number of distinct credential sets plus ten percent of
 headroom for credentials that rotate, never below 8. The result is logged at
-startup. The provider does not start if its ProviderConfigs cannot be listed.
+startup. The provider does not start if its ProviderConfigs and their
+credentials cannot be read within 300 seconds.
 
 The bound is fixed for the lifetime of the process, so ProviderConfigs created
 after startup share the headroom until the next restart. The cache reports
--- a/cmd/provider/cachesize.go
+++ b/cmd/provider/cachesize.go
@@ -1,0 +1,94 @@
+/*
+Copyright 2026 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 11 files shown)


erhancagirici (approved):
@jonasz-lasut thanks for this PR! This strictly improves the fix at #543. LGTM.

Further non-blocking improvement could be a stricter GC on inactive clients, but for now the current shape is enough IMHO.

PR #562, opened 2026-09-03
feat: add noRequeue option to Readiness to suppress periodic poll

### LINKED ISSUE #561: add an option to skip periodic poll for stable resources (SuccessfulCreate + ResourceUpToDate)
In large fleets, provider-kubernetes generates a continuous stream of reconcile events even for resources that have been stable since creation. Every resource is requeued after each poll interval and `Observe()` contacts the remote cluster regardless of whether the resource has ever drifted.

For resources with `readiness.policy: SuccessfulCreate`, the provider already knows after the first successful create that the resource is ready. On every subsequent poll, `Observe()` returns ResourceUpToDate: true unconditionally (no drift is ever detected), yet the resource is still requeued at the configured --poll interval. That makes sense when you want to ensure the resource is still there, but **there should be some sort of fire-and-forget mechanism that can be enabled after the resource is ready at least once** .

At scale this is significant. We have seen a case, for example, with ~8k Object resources × 1 poll/30min ≈ 4 unnecessary remote Kubernetes API calls per second, sustained indefinitely, generating noise in both the control plane cluster (Crossplane events) and the remote clusters (API server load).

Notice this isn't the same as setting `crossplane.io/paused: "true"`. When paused is set, it will also stop receiving updates, meaning any change in the object won't be propagated. What we actually need is an option that stops the `Observe()` calls for resources with `readiness.policy: SuccessfulCreate` once they become ready.

Introduce a new readiness.policy value `SuccessfulCreateFireAndForget` (name open to discussion) that extendsSuccessfulCreate with the additional behaviour of suppressing periodic Observe() calls once the resource is ready.

  Behaviour:
  - First reconcile: behaves identically to SuccessfulCreate. Creates the resource, marks `Ready: True` immediately after the create call succeeds.
  - Subsequent reconciles: skips the remote `Observe()` call entirely and does not requeue periodically.
  - Spec change: the existing watch mechanism triggers an immediate reconcile as normal. The remote `Observe()` runs, and if an update is needed it is applied. Once done, the resource returns to the no-requeue state.

pedromartinssouza:

Adds a `noRequeue` boolean field to the `Readiness` struct (both cluster and namespaced Object types). When set to `true` and the object has `Ready: True`, the `WithPollIntervalHook` returns `0`, suppressing the periodic poll requeue. The provider will not contact the remote cluster again until a spec change triggers a reconcile via the watch mechanism.

This is a pure opt-in performance knob — default is `false`, preserving all existing behaviour. It is orthogonal to `policy` and works with any readiness policy.

Example usage:
```yaml
readiness:
  policy: SuccessfulCreate
  noRequeue: true
```

Fixes #561

I have:

- Unit tests added for `buildPollIntervalHook` in both `internal/controller/cluster/object/object_test.go` and `internal/controller/namespaced/object/object_test.go`, covering all four combinations of `noRequeue` and `Ready` state.
- `go test ./...` passes cleanly.
- CRD YAML regenerated via `go generate ./apis/...`.

### DIFF
--- a/apis/cluster/object/v1alpha2/types.go
+++ b/apis/cluster/object/v1alpha2/types.go
@@ -151,6 +151,14 @@ type Readiness struct {
 	//  `object.status.conditions.all(x, x.status == "True")` mimics the behavior of the AllTrue readiness policy
 	//  `object.status.conditions.exists(c, c.type == "condition1" && c.status == "True" )` checks just one condition
 	CelQuery string `json:"celQuery,omitempty"`
+
+	// NoRequeue suppresses the periodic poll requeue once the object is ready and up-to-date.
+	// The object will still be reconciled immediately on any spec change via the watch mechanism.
+	// This is useful for large fleets of stable resources where periodic Observe() calls to the
+	// remote cluster are unnecessary overhead.
+	// +optional
+	// +kubebuilder:default=false
+	NoRequeue bool `json:"noRequeue,omitempty"`
 }
 
 // ConnectionDetail represents an entry in the connection secret for an Object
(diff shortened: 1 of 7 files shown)


jonasz-lasut:
As discussed in the issue - superseded by crossplane-runtime >=v2.3.0 support for per resource poll interval customization via annotation

PR #569, opened 2026-09-11
Promote Watches feature to beta

### LINKED ISSUE #410: Promote Watches feature to beta
This issue tracks promoting the currently-alpha "watches" feature to beta.

jonasz-lasut:
Promotes Watches feature to beta
Fixes #410

I've kept the default flag value as `false` as both the behavior and the resource consumption of the provider changes with this feature enabled, I'm happy to change it to `true` as it's a beta feature now but I'd like to get a second opinion
@bobh66 what do you think as you've opened the #410 ?

I have:

### DIFF
--- a/cmd/provider/main.go
+++ b/cmd/provider/main.go
@@ -272,8 +272,8 @@ func main() {
 	}
 
 	if *enableWatches {
-		o.Features.Enable(features.EnableAlphaWatches)
-		log.Info("Alpha feature enabled", "flag", features.EnableAlphaWatches)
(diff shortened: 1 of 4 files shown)


bobh66 (commented):
Thanks @jonasz-lasut  - I think in beta it should be enabled by default, especially considering that there is no functionality change unless the `watch` flag in the `Object` is set to `true`.

jonasz-lasut:
Sounds good to me @bobh66 
Pushed a commit that changes the default to `true`

bobh66 (approved):
LGTM - thanks!

PR #571, opened 2026-09-22
Update crossplane-runtime to consume https://github.com/crossplane/crossplane-runtime/pull/1130

### LINKED ISSUE #542: Provider creates a resource even when the Create policy is not included
### What happened?
When defining an `Object` with `managementPolicies: ["Observe", "Update"]` the provider creates the resource if it doesn't exist.

The managed reconciler does not gate the Update() call on observation.ResourceExists [here](https://github.com/crossplane/crossplane-runtime/blob/main/pkg/reconciler/managed/reconciler.go#L1530) so the reconciler will call Observe() for a kubernetes resource that doesn't exist and get back observation.ResourceExists == false and observation.ResourceUpToDate == false, will skip the Create() step as intended, but will call Update() because the resource is not up to date and provider-kubernetes creates the resource since it uses SSA.

This probably needs to be fixed in crossplane-runtime but I wanted to also document it here for information purposes.

### How can we reproduce it?
Sample manifest:

```
kind: Object
metadata:
  name: sample-namespace
spec:
  forProvider:
    manifest:
      apiVersion: v1
      kind: Namespace
  managementPolicies:
    - Observe
    - Update
  providerConfigRef:
    name: default
```

1 - Verify that the "sample-namespace" Namespace does not exist in the cluster
2 - Apply the manifest
3 - Observe that the "sample-namespace" Namespace now exists even though the `Object` does not have the `Create` policy defined.

Crossplane version: 2.3.4

* Cloud provider or hardware configuration: kind
* Kubernetes version (use `kubectl version`): 1.36
* Kubernetes distribution: kind

jonasz-lasut:
https://github.com/crossplane/crossplane-runtime/pull/1130 makes that Update/LateInitialize calls are omitted for not-existing resources, which is crucial for provider-kubernetes, which uses SSA
I've updated crossplane-runtime to latest `main` as we were already importing from main to consume #1113 in crossplane-runtime

Fixes #542

I have:

E2E on ci


bobh66 (approved):
LGTM - thanks @jonasz-lasut

